PR #186, opened 2014-11-25
Clean up and speed up ref attribute resolution

treppo:
For my [schema validation benchmark](https://github.com/treppo/json-schema/blob/benchmarkin/bench/benchmark.rb), this change means a ~10% speedup. I extracted a few methods on the way to clean up.


treppo:
I will close this, as I think that we need a real benchmark tool, so that performance regressions in pull requests are caught. I opened an issue for discussion on this (#218).

PR #189, opened 2014-11-26
Added a require_all option for validating closer to spec

ninkibah:
Added a require_all option that only validates when all properties in the schema are in the data.

It defaults to true, so the current strict mode behavior is unchanged.

It only changes the behavior of draft4 type schemas. Sorry, I don't know enough to know if I should change the other schemas.

Discussion was started on this issue: https://github.com/ruby-json-schema/json-schema/issues/103

### DIFF
--- a/VERSION.yml
+++ b/VERSION.yml
@@ -1,3 +1,3 @@
 major: 2
 minor: 4
-patch: 1
+patch: 1.1
--- a/lib/json-schema/attributes/properties.rb
+++ b/lib/json-schema/attributes/properties.rb
@@ -67,7 +67,7 @@ module JSON
       # draft4 relies on its own RequiredAttribute validation at a higher level, rather than
       # as an attribute of individual properties.
       def self.required?(schema, options)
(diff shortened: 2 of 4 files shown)


RST-J:
If we consider to take this in, I think ```require_all``` should default to ```false```. Why would it be closer to the spec? The spec, from draft3 on says, that property definitions are optional unless stated as required.

iainbeeston:
I haven't been keeping up with draft 5 but I think there might be something
like this in there. Maybe.

My personal preference would be keeping to what's in the spec, to keep the
complexity down...

RST-J:
I agree and would rather remove strict for v3 than add another non-spec mode. Again, if people want that, it should be another gem, maybe as some kind of plugin - which preferably adds some kind of preprocessor that modifies the given schema such that everything is required. This would also work for things like stringification of symbols and inserting defaults plus there would be vanilla json-schema as well.
I'll close this as we are already aiming for v3.

PR #205, opened 2014-12-02
Stop running 1.8 on travis

pd:
There is likely some 1.8-specific code juggling that still remains, but I think this is sufficient for now: we explicitly aren't testing against 1.8 any longer.

I've added a Changelog.md, but didn't put any real thought into its structure or such yet. I figure once there's more than a single line we can re-visit that. =)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: "ruby"
 
 rvm:
-  - "1.8"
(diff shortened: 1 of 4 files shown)


RST-J:
So this should be fine now, isn't it?

iainbeeston:
Yep! :+1:

Issue #195, opened 2014-11-29
Replace yajl and json gem with multi JSON?


iainbeeston:
Right now there is a bit of custom code for both yajl and json-gem, which both behave slightly differently. I'd like to start a discussion about what the future should be for that.

I'd suggest that either:
1. We switch to using multi_json (which supports all manner of json parsers with a unified api), or
2. We switch to only supporting json (like rails did) and simplify the code a bit

There is an argument that [multi_json is deprecated](https://github.com/intridea/multi_json/pull/113) and nowadays the right move is for everyone to use json-gem, but I can't help thinking people should be given the choice (and therefore multi json is better)

flower-pot:
In that case, only the code that specialises on yajl has to be removed if I am correct? I would do the changes if you can confirm that it is the consensus.

iainbeeston:
Yes, I think so

flower-pot:
Nevermind, it seems like #206 already did the work, I looked at the wrong branch. I guess this issue can be closed?

iainbeeston:
Sorry, yes. It's schedule for v3.

PR #206, opened 2014-12-02
Remove support for alternate JSON backends

pd:
See #195; there wasn't broad agreement there, but if you look at the multi_json PR that @iainbeeston linked (https://github.com/intridea/multi_json/pull/113), you can see just how broadly this is happening across the ruby ecosystem right now. With the dropping of support for 1.8 (#205), we have a built-in, not-that-bad JSON library on all supported runtimes, so we should just use that, IMO.

This will have conflicts with #205, since both add a Changelog.md. I assume that one will land first, so I'll rebase later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,6 @@ install:
 matrix:
   include:
     - rvm: "2.1"
-      gemfile: "gemfiles/Gemfile.multi_json.x"
-    - rvm: "2.1"
-      gemfile: "gemfiles/Gemfile.yajl-ruby.x"
-    - rvm: "2.1"
       gemfile: "gemfiles/Gemfile.uuidtools.x"
   allow_failures:
     - rvm: "1.8"
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,0 +1,4 @@
+# Changes in json-schema 3.0
+
+* Support for `multi_json` and `yajl-ruby` have been dropped; the standard `json` library,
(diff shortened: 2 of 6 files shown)


iainbeeston:
What else do we need to do to get this merged?

Are there any other references to other json backends?

.travis.yml
```diff
@@ -16,10 +16,6 @@ install:
 matrix:
   include:
     - rvm: "2.1"
-      gemfile: "gemfiles/Gemfile.multi_json.x"
```
iainbeeston: Can we remove these extra gem files as well?

RST-J:
I don't know whether there is anything more to do, but as the tests pass I guess there should be nothing left. :+1:

iainbeeston:
I think we should remove the alternate gemfiles (that test different json backends) before merging this.

iainbeeston:
I suggest we also add a changelog entry for this (see #232 )

RST-J:
@hoxworth could you live with that?

hoxworth:
Compatibility mode is fine for the alternate backends, we can move forward on this. :+1:

iainbeeston:
I've just rebased the 3.x branch to include all the latest changes from master.

However, there's a merge conflict on this branch so I can't merge it

iainbeeston:
I've resolved the conflict manually and merged in a29c0526a7d710319decc827dc7f21a1187389b9

PR #209, opened 2014-12-02
Codeclimate code coverage

iainbeeston:
This will start sending test coverage data to code climate, so you can see it inline, alongside code quality metrics


iainbeeston:
Merging #206 will make this work, otherwise I'll need to explicitly specify which backend we're using

iainbeeston:
Closing this in favour of #209

PR #213, opened 2014-12-16
End single quote in validation error

olleolleolle:
A trivial change to clarify the output message for `contained undefined properties`: add the closing single quote.


iainbeeston:
Merging this... Good, catch, thanks!

PR #216, opened 2015-01-15
Show sub-errors for oneOf

isage:
anyOf and allOf show sub-errors.
with this patch oneOf do the same

### DIFF
--- a/lib/json-schema/attributes/oneof.rb
+++ b/lib/json-schema/attributes/oneof.rb
@@ -12,23 +12,26 @@ module JSON
         original_data = data.is_a?(Hash) ? data.clone : data
         success_data = nil
 
+        valid = false
+
         one_of.each_with_index do |element, schema_index|
           schema = JSON::Schema.new(element,current_schema.uri,validator)
           pre_validation_error_count = validation_errors(processor).count
           begin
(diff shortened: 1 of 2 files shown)


iainbeeston:
Great idea, could you add some tests for this similar to those for any of and all of?
(e.g. in test_all_of_ref_schema.rb and test_any_of_ref_schema.rb)

iainbeeston:
Also, this change seems to cause a test failure related to oneOf

isage:
Looking at the test, i'm not sure why it was working before.
insert_defaults will insert foo => "view" into data, so data now matches both variants of oneOf, which should fail.

But still, there's seems to be an error in my patch, because even {foo => view} doesn't seem to validate
I'll try to fix this and write tests

isage:
Oh, okay, i finally get, how test works. "default" inserts value only for one oneOf case, which will fail with strict, and the first one will pass.
I forgot to handle exceptions, that's why it was failing. Now it works.
I'll write tests for oneOf tomorrow

isage:
Okay, i've added tests for sub-errors

RST-J:
Seems fine to me :+1:

iainbeeston:
Ok, I'm happy with this :+1: 

Merging

PR #219, opened 2015-01-20
Issue with Fixnum and Float in enum

RST-J:
I found out that there is a problem with Fixnum und Float values not being treated as equal when checking against an enum list, which they should if the required type is number as it includes all integers.

I don't know why ```include?``` fails here.

And another thing. There are currently two tests failing for draft-1 and draft-2 when they are called with schema validation. The error is: ```JSON::Schema::SchemaError: Schema not found: http://json-schema.org/draft-02/hyper-schema#```
Any quick hints how to fix that? I'm in the office and cannot dig into that right now.

### DIFF
--- a/test/support/enum_validation.rb
+++ b/test/support/enum_validation.rb
@@ -1,0 +1,76 @@
+module EnumValidation
+  def test_enum
+    schema = {
+      "properties" => {
+        "a" => {"enum" => [1,'boo',[1,2,3],{"a" => "b"}]}
+      }
+    }
+
+    data = { "a" => 1 }
+    assert_valid schema, data
+
+    data["a"] = 'boo'
+    assert_valid schema, data
+
+    data["a"] = [1,2,3]
+    assert_valid schema, data
+
+    data["a"] = {"a" => "b"}
+    assert_valid schema, data
+
+    data["a"] = 'taco'
+    refute_valid schema, data
+
+    data = {}
+    assert_valid schema, data
+  end
+
+  def test_fixed_float_issue_enum
+    schema = {
+      "properties" => {
+        "a" => {
+          "type" => "number",
+          "enum" => [0, 1, 2]
+        }
+      }
+    }
+
+    data = { "a" => 0 }
+    assert_valid schema, data
+
+    data["a"] = 0.0
+    assert_valid schema, data
+
+    data["a"] = 1
+    assert_valid schema, data
+
+    data["a"] = 1.0
+    assert_valid schema, data
+  end
+
+  def test_enum_with_schema_validation
+    schema = {
+      "properties" => {
+        "a" => {"enum" => [1,'boo',[1,2,3],{"a" => "b"}]}
+      }
+    }
+    data = { "a" => 1 }
+    assert_valid(schema, data, :validate_schema => true)
+  end
+
+  module ItemsTests
+    def test_items_single_schema
+      schema = { 'items' => { 'type' => 'string' } }
+
+      assert_valid schema, []
+      assert_valid schema, ['a']
+      assert_valid schema, ['a', 'b']
+
+      refute_valid schema, [1]
+      refute_valid schema, ['a', 1]
+
+      # other types are disregarded
+      assert_valid schema, {'a' => 'foo'}
+    end
+  end
+end
--- a/test/support/enum_validation.rb
+++ b/test/support/enum_validation.rb
@@ -1,5 +1,5 @@
 module EnumValidation
-  def test_enum
+  def test_enum_general
     schema = {
       "properties" => {
         "a" => {"enum" => [1,'boo',[1,2,3],{"a" => "b"}]}
@@ -25,7 +25,7 @@ module EnumValidation
     assert_valid schema, data
   end
 
-  def test_fixed_float_issue_enum
+  def test_enum_number_integer_includes_float
     schema = {
       "properties" => {
         "a" => {
@@ -48,6 +48,54 @@ module EnumValidation
     assert_valid schema, data
   end
 
+  def test_enum_number_float_includes_integer
+    schema = {
+      "properties" => {
+        "a" => {
+          "type" => "number",
+          "enum" => [0.0, 1.0, 2.0]
+        }
+      }
+    }
+
+    data = { "a" => 0.0 }
+    assert_valid schema, data
+
+    data["a"] = 0
+    assert_valid schema, data
+
+
+    data["a"] = 1.0
+    assert_valid schema, data
+
+    data["a"] = 1
+    assert_valid schema, data
+  end
+
+  def test_enum_integer_integer_excludes_float
+    schema = {
+      "properties" => {
+        "a" => {
+          "type" => "integer",
+          "enum" => [0, 1, 2]
+        }
+      }
+    }
+
+    data = { "a" => 0 }
+    assert_valid schema, data
+
+    data["a"] = 0.0
+    refute_valid schema, data
+
+
+    data["a"] = 1
+    assert_valid schema, data
+
+    data["a"] = 1.0
+    refute_valid schema, data
+  end
+
   def test_enum_with_schema_validation
     schema = {
       "properties" => {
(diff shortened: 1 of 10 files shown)


RST-J:
I fixed that by simply adding hyperschema validators for draft1 and 2 according to the one already present for draft4. Could that have done any harm? And I have no idea why it is not required for draft3.

Another thing which occurred to me: Because I renamed some test, it failed. That was because that test was overwritten by another one with the same method name. Is there a way we can Minitest have raising an exception when a test method is about to be redefined in a test class?

iainbeeston:
I don't think minitest or test unit would be able to raise an exception if a test is redefined, because it's just a plain old Ruby object. I imagine the best we can do is to make sure tests are run with warnings enabled. I've just raised a pull request to do that

test/support/enum_validation.rb
```diff
@@ -0,0 +1,133 @@
...
+
+
+      data["a"] = 1.0
+      assert_valid schema, data
+
+      data["a"] = 1
+      assert_valid schema, data
+    end
+
+    def test_enum_integer_integer_excludes_float
```
iainbeeston: Should this be test_enum_integer_excludes_float ?

iainbeeston:
All sounds fine to me. I'm not sure why any? works but include does not (if it's just an array) but I'm not too concerned

:+1:

iainbeeston:
I've just realised why the change from include? to any? works. In the way you have defined any, we check "data == value" (where value is an element in the array). Ruby's Array#include? method checks "value == data". By reversing the order we change which object's == method is called. It looks like whatever data#== is it is not behaving the same as value#==.

Ideally == should be symmetric and the two should be equivalent. Perhaps a better fix to this bug would be to correct data#== (but I have no idea what class data is in this instance)

RST-J:
That seems not to be the cause, because the tests still pass when I use ```enum.any? { |value| value == data }```.
When I started to go for this I also found out that, although ```enum.include?(data)``` did not work, calling ```.include?(data)``` on a literal definition of enum worked in the debugger session.

iainbeeston:
Do you have any idea what class enum is?

RST-J:
Good one @iainbeeston :-), the problem is not the type of the values but of ``enum`` which is [ArraySet](https://github.com/ruby-json-schema/json-schema/blob/master/lib/json-schema/util/array_set.rb), a custom class aiming for improved lookup-speed which fails in this case because ``1.0`` is just equal to ``1`` but not identical.

I don't know what the performance impact for this change is. The current fix would make it obsolete, so we could just go back to use an array here and ``include?`` will work fine again.
If its a noteworthy improvement, I would change the creation of the hash in ``ArraySet.include?`` to add the integer/float variant for equal values.
@hoxworth can you tell something more on this and suggest which way to follow?

iainbeeston:
It might be nice to remove arrayset completely in 3.0. When json-schema was
written it had to work with ruby 1.8, where hashes are unordered. If we
drop 1.8 support in 3.0 then we could potentially replace ArraySet with a
simple hash (which can emulate an ordered set/bag)

iainbeeston:
Actually we don't even need ordering. So a Set would do. Which goes back to
@hoxworth really, and why we need ArraySet (rather than using a Set)

RST-J:
``Set`` is also part of 1.8.7 and doesn't json-schema require at least Ruby 1.8.7 since some releases? So we could use it right away.

But that still wouldn't fix the issue here because a set containing 1 responds with ``false`` when asked whether 1.0 is included (and the other way around).
So either way we have to add equal ``Fixnum``/``Float`` values for the other if we use some sort of set.

iainbeeston:
Oh I see. Set#include? relies on Hash#include?, which checks inclusion
based on #hash,  1.0.hash is not the same as 1.hash, so Set#include?
doesn't work as we expect.

Set supposedly implements enumerable, which explicitly checks inclusion
using ==. So either the ruby documentation is wrong or there's a bug in
ruby.

RST-J:
The docs for [Set](http://www.ruby-doc.org/stdlib-2.1.0/libdoc/set/rdoc/Set.html) explicitly state how equality is handled:
> Equality of elements is determined according to Object#eql? and Object#hash.
``1.eql?(1.0)`` yields ``false``, what btw is how I (intuitively) would expect a set to behave.

iainbeeston:
Set implements Enumerable, which states:

Returns true if any member of enum equals obj. Equality is tested using ==.

http://ruby-doc.org/core-2.2.0/Enumerable.html#method-i-include-3F

RST-J:
Then we should probably submit this contradiction somwhere&trade;. Where is the place for such a thing?

And we need a decision for json-schema for which I'd like to wait what @hoxworth has to say.

iainbeeston:
I'll file a pull request to ruby later. Generally it seems that if the
implementation contradicts the documentation, the docs are wrong.

hoxworth:
In this case I feel like we should treat everything in an enum and all comparisons in the enum as a `Float`, converting `Fixnum` to float in both the insert and lookup case. I honestly can't even remember why `ArraySet` was created instead of `Set`, and it was only a year ago... I'm perfectly fine with moving towards `Set`.

RST-J:
@hoxworth, probably because otherwise schema validation fails with the following message?
> JSON::Schema::ValidationError: The property '#/properties/a/enum' of type Set did not match the following type: array

hoxworth:
@RST-J Ah yes, that is exactly why - `ArraySet` was introduced to fix that exact bug.

iainbeeston:
I've been thinking about this. Array#include? returns true if any element
== the argument. ArraySet#include? returns true if any element eql? the
argument. If ArraySet is supposed to be a drop-in replacement for Array,
and is mainly used for it's include? method, then should that method
really behave differently?

I would argue not.

Perhaps a safer way to optimise include? would be to store a sorted version
of the array in an instance variable (where we now store a set) then
performing a binary search, but we'd have to implement the binary search
ourselves (bsearch is not in Ruby 1.8 or 1.9)

RST-J:
I agree, it should behave like Array and that is what it does now (unless there is another case we haven't figured out yet).

Using a binary search won't work because we cannot order the set with reasonable effort in the general case. The spec says:
> Elements in the array MAY be of any type, including null. 
So we'd need to come up with a function that assigns equal arguments the same value.
This would imply a lot of potential for flaws (only occurring in their special circumstances) and I'm not sure whether the overhead of this function would eat up the benefits of binary search. Maybe ``hash`` would be a solution, but how unique is that?

Do you have a serious concern about my last change, @iainbeeston?

iainbeeston:
I think what you've committed is the best we can do right now. The only
feedback I have on that is:

1. Should we require set in that file? (I assume it's already required
elsewhere)
2. I'd recommend extracting out the "if is_a?(Fixnum) then to_f" to a
method (which normalises values) that can be called twice. I always find
that find of thing makes intent clearer, and makes it obvious that we're
doing the same thing to all values under conlarison.

Otherwise it's :+1: from me

RST-J:
I didn't find any require statement for Set with grep, I don't know why it worked. But I added a require statement for consistency.
And I agree about extracting a function, done.

iainbeeston:
FYI I've updated ruby's rdoc for set to make it explicit that it doesn't use == for include? https://github.com/ruby/ruby/pull/831

inglesp:
Looks mostly fine.  One question:  why do the new hyperschemas include `JSON::Validator.register_default_validator(self.new)`?  As there's only ever going to be one default validator, wouldn't it make sense to set it explicitly?

RST-J:
I didn't spent thoughts here, just copied hyper schema validator for [draft4](https://github.com/ruby-json-schema/json-schema/blob/master/lib/json-schema/validators/hyper-draft4.rb).

I think, generally you are right, the default validator should only be set once and it should be draft4. So the right step is to not do it in the two ne hyper schema validators - I'll remove that - and also not doing it in the hyper-schema-draft 4. I'll prepare another PR to remove that.

RST-J:
I removed all default registrations, also for draft4 in the extra commit.

inglesp:
Good stuff RST-J!

hoxworth:
:+1:

PR #221, opened 2015-01-22
Renamed variable in oneOf to avoid name clash

iainbeeston:
I noticed in #216 that we have some bad variable naming. In the OneOf attribute, there's a `validation_errors` instance variable, but the parent (Attribute) class defines a `validation_errors()` method. So in the same file we have both a method and a variable with the same name. This seems to work, but it's confusing and seems like a bad idea.

This renames that variable to `validation_error_count`, to avoid confusion

### DIFF
--- a/lib/json-schema/attributes/oneof.rb
+++ b/lib/json-schema/attributes/oneof.rb
@@ -6,7 +6,7 @@ module JSON
       def self.validate(current_schema, data, fragments, processor, validator, options = {})
         errors = Hash.new { |hsh, k| hsh[k] = [] }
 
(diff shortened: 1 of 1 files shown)


RST-J:
:+1:

inglesp:
Looks good -- although Ruby is smart enough to distinguish between local variables and instance methods.

iainbeeston:
My intention is to make the code easier to understand, and to get rid of
warning messages in Ruby and linters like rubocop. Functionally it makes no
difference

hoxworth:
:+1:

Issue #32, opened 2012-03-29
Is there a method available to get a fully resolved schema as a hash?


bruno-c:
Is it possible to retrieve the schema that is constructed internally for validation, including extends, $ref'ed schemas, etc as one fully 'resolved' json-schema hash?

I did not have the time to investigate the internals of the gem yet so I thought i'd ask first. I wrote a half-assed implementation myself for my purposes (documentation of the schema) but I figured this might be something that would be useful to others as well.

Thank you.

PR #222, opened 2015-01-24
WIP: resolve refs in schemas in one extrapolated hash

RST-J:
This is a first shot at generating a fully resolved schema hash as requested in #32.
I commented very verbosely so you can directly follow my ideas. Any comments, improvements et al?

Notes:
- So far this only works with $ref but can be easily extended for extends too
- in the test `test_linked_list_ref` using the file path instead of the URL does not work (see comment)
- pretty printing a cyclic hash or using it for validation kills kittens

Questions:
1. Has someone an idea why using the file path in the test does not work?
2. Is there anything missing to be draft4 compliant?

### DIFF
--- a/lib/json-schema/validator.rb
+++ b/lib/json-schema/validator.rb
@@ -137,6 +137,10 @@ module JSON
       build_schemas(schema)
     end
 
+    def base_schema
+      @base_schema
+    end
+
     def absolutize_ref_uri(ref, parent_schema_uri)
       ref_uri = Addressable::URI.parse(ref)
 
@@ -238,6 +242,46 @@ module JSON
       @errors
     end
 
+    def resolve_nested_references(schema, schema_portion, list)
+      case schema_portion
+      when Array
+        # for arrays resolve all items which have not been resolved yet
+        schema_portion.each_with_index do |item, i|
+          schema_portion[i] = resolve_nested_references(schema, item, list) unless list.include?(schema_portion[i])
+        end
+      when Hash
+        if schema_portion["$ref"]
+          # for hashes with a reference, resolve their referenced schema and copy(!) the result into the currently handled portion
+          _, ref_schema = JSON::Schema::RefAttribute.get_referenced_uri_and_schema(schema_portion, schema, self)
+          resolved_portion = resolve(ref_schema, ref_schema.schema, list)
+          schema_portion.clear
+          resolved_portion.each { |key, value| schema_portion[key] = value }
+        else
+          # for hashes without a reference we also resolve recursively what has not been resolved yet
+          schema_portion.each do |key, value|
+            schema_portion[key] = resolve_nested_references(schema, value, list) unless list.include?(value)
+          end
+        end
+      end
+      schema_portion
+    end
+
+    def resolve(schema, handle_schema_hash, list = [])
+      # if we already dealt with the schema hash we have nothing left to do
+      return handle_schema_hash if list.include?(handle_schema_hash)
+      # remember the hash we are dealing with for the future
+      list << handle_schema_hash
+      # $ref is specified as 'SHOULD replace the current schema with the schema referenced by the value's URI' so we do exactly that
+      if handle_schema_hash["$ref"]
+        _, ref_schema = JSON::Schema::RefAttribute.get_referenced_uri_and_schema(handle_schema_hash, schema, self)
+        fail "Could not find referenced schema #{handle_schema_hash["$ref"]}" unless ref_schema
+        # translates to 'replace the current schema with the referenced one and resolve this instead'
+        handle_schema_hash = resolve(ref_schema, ref_schema.schema, list)
+      end
+      # finally we must recursively walk through our schema and resolve nested schemas
+      resolve_nested_references schema, handle_schema_hash, list
+      handle_schema_hash
+    end
 
     class << self
       def validate(schema, data,opts={})
@@ -277,6 +321,11 @@ module JSON
         opts[:record_errors] = true
         validator = JSON::Validator.new(schema, data, opts)
         validator.validate
+      end
+
+      def fully_resolve(schema)
+        validator = JSON::Validator.new(schema, {}, {})
+        validator.resolve(validator.base_schema, validator.base_schema.schema)
       end
 
       def fully_validate_schema(schema, opts={})
--- a/test/test_fully_resolve.rb
+++ b/test/test_fully_resolve.rb
@@ -1,0 +1,111 @@
+# encoding: utf-8
+require File.expand_path('../test_helper', __FILE__)
+
+class FullyResovleDraft4Test < Minitest::Test
+  def test_top_level_ref
+    schema = {
+      "$schema" => "http://json-schema.org/draft-04/schema#",
+      "$ref" => "test/schemas/fully_resolve/leave_string_schema.json#"
+    }
+
+    expected = {
+      "$schema" => "http://json-schema.org/draft-04/schema#",
+      "type" => "object",
+      "required" => ["name"],
+      "properties"  => {
+        "name"  => { "type" => "string" }
+      }
+    }
+
+    resolved = JSON::Validator.fully_resolve(schema)
+    assert_equal(resolved, expected, "should resolve a top level $ref reference")
+
+    # Test that the result is usable for validation
+    data = {name: '2'}
+    assert_valid(schema, data)
+    assert_valid(resolved, data)
+  end
+
+  def test_one_of_ref
+    schema = {
+      "$schema" => "http://json-schema.org/draft-04/schema#",
+      "oneOf" => [
+        { "$ref" => "test/schemas/fully_resolve/leave_string_schema.json#" },
+        { "$ref" => "test/schemas/fully_resolve/leave_integer_schema.json#" }
+      ]
+    }
+
+    expected = {
+      "$schema" => "http://json-schema.org/draft-04/schema#",
+      "oneOf" => [
+        {
+          "$schema" => "http://json-schema.org/draft-04/schema#",
+          "type" => "object",
+          "required" => ["name"],
+          "properties"  => {
+            "name"  => { "type" => "string" }
+          }
+        },
+        {
+          "$schema" => "http://json-schema.org/draft-04/schema#",
+          "type" => "object",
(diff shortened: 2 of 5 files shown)


lib/json-schema/validator.rb
```diff
@@ -137,6 +137,10 @@ def load_ref_schema(parent_schema, ref)
       build_schemas(schema)
     end
 
+    def base_schema
```
inglesp: It's normal to use `attr_accessor` for this kind of thing.

lib/json-schema/validator.rb
```diff
@@ -238,6 +242,46 @@ def validation_errors
       @errors
     end
 
+    def resolve_nested_references(schema, schema_portion, list)
```
inglesp: `schema_portion` appears to be a name that's new to the project.  Could you pick a name that's already been used before, so it's easier to work out what it refers to?

Also, it's not clear what `list` is.  Could you find a more suitable name?

lib/json-schema/validator.rb
```diff
@@ -238,6 +242,46 @@ def validation_errors
...
+          # for hashes without a reference we also resolve recursively what has not been resolved yet
+          schema_portion.each do |key, value|
+            schema_portion[key] = resolve_nested_references(schema, value, list) unless list.include?(value)
+          end
+        end
+      end
+      schema_portion
+    end
+
+    def resolve(schema, handle_schema_hash, list = [])
```
inglesp: It's not clear to me what `handle_schema_hash` is, and again, I think that `list` could have a better name.

lib/json-schema/validator.rb
```diff
@@ -238,6 +242,46 @@ def validation_errors
...
+      list << handle_schema_hash
+      # $ref is specified as 'SHOULD replace the current schema with the schema referenced by the value's URI' so we do exactly that
+      if handle_schema_hash["$ref"]
+        _, ref_schema = JSON::Schema::RefAttribute.get_referenced_uri_and_schema(handle_schema_hash, schema, self)
+        fail "Could not find referenced schema #{handle_schema_hash["$ref"]}" unless ref_schema
+        # translates to 'replace the current schema with the referenced one and resolve this instead'
+        handle_schema_hash = resolve(ref_schema, ref_schema.schema, list)
+      end
+      # finally we must recursively walk through our schema and resolve nested schemas
+      resolve_nested_references schema, handle_schema_hash, list
```
inglesp: Could you be consistent about calling functions with parentheses?

test/test_fully_resolve.rb
```diff
@@ -0,0 +1,282 @@
+# encoding: utf-8
+require File.expand_path('../test_helper', __FILE__)
+
+class FullyResovleDraft4Test < Minitest::Test
+  def test_top_level_ref
+    schema = {
+      "$schema" => "http://json-schema.org/draft-04/schema#",
+      "$ref" => "test/schemas/fully_resolve/leave_string_schema.json#"
```
inglesp: Do you mean `leaf` instead of `leave`?

test/test_fully_resolve.rb
```diff
@@ -0,0 +1,282 @@
...
+    assert_valid(resolved, data)
+  end
+
+  def test_linked_list_ref
+    stub_request(:get, "example.com/linked_list").to_return(:body => File.new('test/schemas/fully_resolve/linked_list_schema.json'), :status => 200)
+
+    schema = {
+      "$schema" => "http://json-schema.org/draft-04/schema#",
+      "$ref" => "http://example.com/linked_list#"
+      # it does not work to reference the schema by path
```
inglesp: Why not?

test/test_fully_resolve.rb
```diff
@@ -0,0 +1,282 @@
...
+      "type" => "array",
+      "items" => { "type" => "string" },
+      "minItems" => 1,
+      "uniqueItems" => true
+    }
+    simple_types = {
+      "enum" => [ "array", "boolean", "integer", "null", "number", "object", "string" ]
+    }
+    expected = {}
+    expected.merge!({
```
inglesp: You don't need to create an empty hash and then `merge!`, do you?

inglesp:
I've had a look at this because it's a feature that I'd like to see in the gem.  I think the approach is sensible, but I don't know the library very well so there might be some edge cases I'm missing.

I've made a handful of stylistic comments which I hope are helpful!

lib/json-schema/validator.rb
RST-J: I'll check whether that's how calls are made elsewhere in the lib. In my company we use parentheses in nested calls but omit them if the hole expression is only a function call. But I'm not particularly tied to this, so I can change this.

test/test_fully_resolve.rb
RST-J: As mentioned in the PR message, unfortunately I don't know. I asked because I hope @pd or @iainbeeston have an idea what's going wrong here.

test/test_fully_resolve.rb
RST-J: I do, because I need a reference to the outer hash within the merged hash which describes the resolved schema as it is self-referencing.

RST-J:
Thanks for your feedback. I'll see what I can do about the names and particularly ```list``` is really not self describing.

lib/json-schema/validator.rb
RST-J: The only function which receives the same combination, a schema object and a hash describing some property (or the entire) of the schema is in [get_referenced_uri_and_schema](https://github.com/ruby-json-schema/json-schema/blob/master/lib/json-schema/attributes/ref.rb#L21) in ```RefAttribute``` where the hash, here ```schema_portion``` is called ```s```.
What do you think of ```schema_hash``` as name? Or ```schema_portion_hash``` as it can also be some nested sub-property of the schema?

lib/json-schema/validator.rb
iainbeeston: I'm not aware of any style guide for json-schema but maybe we should add
rubocop or Hound to check all new code? (I don't feel very strongly about
this - there are bigger fish to fry)

test/test_fully_resolve.rb
iainbeeston: Sorry, I'm not sure, without investigating in depth

lib/json-schema/validator.rb
RST-J: Actually I think there is no better name. Values coming in here as ```schema_portion``` are either Hashes, Arrays or primitive values and they are all nested parts of the schema to be resolved (they are neither all hashes nor all valid schemas on their own).

lib/json-schema/validator.rb
RST-J: I think about calling the parameter just ```schema_hash```. It is the hash representing the schema to be resolved (or subschema in nested calls). Its the primary structure the resolution process is working on. The ```schema``` parameter is an actual ```JSON::Schema``` instance which is required to fetch references.

test/test_fully_resolve.rb
BGarretson: You can ref files on the file system, but they have to live at the same depth or deeper than the 'entrypoint' schema that contains the $ref. Assuming that condition is met, you can reference by filename or folder+filename, followed by a JSON pointer. I have done so with this gem and a multi-file json schema for many months now. 

assuming:

./main.json
./secondary.json
./foo/tertiary.json

a refs inside main.json would look like:

"$ref": "secondary.json#/definitions/something"
"$ref": "foo/tertiary.json#"

This assumes the ID of the main.josn file has been modified to reflect it's absolute location on the filesystem prior to loading, and I'm not sure how this would play with a schema created in-memory referencing a file system location, but if you set up the id correctly to be a file:/// location, it should still work.

BGarretson:
I tested this branch using the following test:

```
# test
    context 'testing schema resolver code' do
        it 'does not throw an error' do
            schema = JSON::Validator.get_schema_body('./schemas/iodocs/v2/expanded/iodocs_schema_v2.0.json')
            resolved = JSON::Validator.fully_resolve(schema)
            json = resolved.to_json(max_nesting: 10000)
            json
        end
    end
```

It generated a NestingError regardless of the depth I allowed in the call above - this is definitely a serialization issue where the object being serialized has infinite depth due to circular references in the way the hashes are related to each other. 

```
JSON::NestingError: nesting of 10000 is too deep
./spec/lib/cli_spec.rb:156:in `to_json'
./spec/lib/cli_spec.rb:156:in `block (3 levels) in <top (required)>'
/Users/bgarrets/.rvm/gems/ruby-2.0.0-p451/gems/ruby-debug-ide-0.4.24/lib/ruby-debug-ide.rb:86:in `debug_load'
/Users/bgarrets/.rvm/gems/ruby-2.0.0-p451/gems/ruby-debug-ide-0.4.24/lib/ruby-debug-ide.rb:86:in `debug_program'
-e:1:in `load'
-e:1:in `<main>'
```

It's not necessarily enough to create a hash in memory and reuse existing hash object references if they can be reused during buildout. In any sufficiently complex schema where objects are being reused in multiple places in the schema hierarchy you'll end up with circular references that can't then be serialized out. 

To get around this, you would need to basically collapse every schema you encounter into the 'new schema file's definitions block, possibly generating new definition entries for schemas that were the root schema in a file, and updating $ref statements to point to the new schema location within the single file schema instead of reaching out to other files.

BGarretson:
The suggestion above is basically the process I had to go through to manually create a single-file schema from the multi-file schema set as a manual workaround for my colleague.

BGarretson:
Alternately, you could monitor for circular references and only allow the method to resolve a schema if it was a directed acyclical graph

RST-J:
I like your first idea better and will have a look at how to do this when I find time.

Jubke:
this has been quiet for a while... any updates or chances it will be finished and merged?

ubhide:
Has this been abandoned?

DeltaRazero:
Any progress on this? This is very trivial feature for a web application that I am working on right now.

bastelfreak:
@DeltaRazero we need to rebase this to merge it. If you're interested you can checkout the branch, rebase against our master and submit it as new PR.

TheMeier:
this is quite old and needs some work. 
close due to inactivity, please feel free to re-open any time

PR #223, opened 2015-01-28
Added Ruby 2.2 to the build matrix

iainbeeston:
Ruby 2.2 came out over a month ago so we should probably be testing against that now


RST-J:
:+1:

inglesp:
Looks good.

hoxworth:
:+1:

PR #224, opened 2015-01-28
Made sure we really do update the common test suite before test runs

iainbeeston:
It seems that the code we have for updating submodules wasn't really
working. So far as I can tell it was not updating the code in the
submodule from the remote repository, and the .gitmodules file was set
to track master, not develop (which is the default on the common test
suite project)


RST-J:
:+1:

inglesp:
Looks good.

hoxworth:
:+1:

PR #226, opened 2015-02-06
Updated to addressable 2.3.7

iainbeeston:
To ensure that invalid uris of the form "::http" throw exceptions rather than failing silently. This aught to make uri parsing more robust

Addressable has finally resolved the bug in https://github.com/ruby-json-schema/json-schema/pull/174 so I suggest we use the fix and hope that it makes uri parsing more robust. I suspect that there should be other changes elsewhere in the code (such as gracefully handling uris in sensitive places).

@RST-J I'd appreciate your feedback on this especially

### DIFF
--- a/json-schema.gemspec
+++ b/json-schema.gemspec
@@ -24,5 +24,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency "webmock"
   s.add_development_dependency "bundler"
 
-  s.add_runtime_dependency "addressable", '~> 2.3'
+  s.add_runtime_dependency "addressable", '~> 2.3.7'
 end
--- a/lib/json-schema/attributes/formats/uri.rb
+++ b/lib/json-schema/attributes/formats/uri.rb
@@ -7,10 +7,7 @@ module JSON
         return unless data.is_a?(String)
         error_message = "The property '#{build_fragment(fragments)}' must be a valid URI"
         begin
-          # TODO
-          # Addressable only throws an exception on to_s for invalid URI strings, although it
-          # probably should throughout parse already - https://github.com/sporkmonger/addressable/issues/177
-          Addressable::URI.parse(data).to_s
+          Addressable::URI.parse(data)
         rescue Addressable::URI::InvalidURIError
           validation_error(processor, error_message, fragments, current_schema, self, options[:record_errors])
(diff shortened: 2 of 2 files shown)


json-schema.gemspec
```diff
@@ -24,5 +24,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency "webmock"
   s.add_development_dependency "bundler"
 
-  s.add_runtime_dependency "addressable", '~> 2.3'
+  s.add_runtime_dependency "addressable", '~> 2.3', '>= 2.3.7'
```
RST-J: Why not directly ``'~> 2.3.7'``, is that not possible or does it have a different meaning?

lib/json-schema/attributes/formats/uri.rb
```diff
@@ -7,9 +7,6 @@ def self.validate(current_schema, data, fragments, processor, validator, options
         return unless data.is_a?(String)
         error_message = "The property '#{build_fragment(fragments)}' must be a valid URI"
         begin
-          # TODO
-          # Addressable only throws an exception on to_s for invalid URI strings, although it
-          # probably should throughout parse already - https://github.com/sporkmonger/addressable/issues/177
           Addressable::URI.parse(data).to_s
```
RST-J: ``Addressable::URI.parse(data)`` should be sufficient here now.

RST-J:
>I suspect that there should be other changes elsewhere in the code (such as gracefully handling uris in sensitive places)
What do you mean by other changes? That we probably have places where we can omit checks that are now handled by addressable? Maybe. But the change here, when the ``to_s`` is removed, is good on its own.

json-schema.gemspec
iainbeeston: No, you're right (I thought that ~> 2.3.7 meant >= 2.3 && < 3.0 - but
apparently not)

iainbeeston:
I see. I didn't have anything specific in mind. I'll remove the to_s

iainbeeston:
@RST-J That's done now

RST-J:
:+1:

inglesp:
Looks good.

hoxworth:
:+1:

PR #231, opened 2015-02-10
Enabled warnings when running tests

iainbeeston:
As mentioned in #219 we want to avoid issues such as redefined methods. Running tests with warnings enabled should help.

Note that this shows up warnings for some existing issues, including some in addressable, so I recommend we merge #226 before this

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -14,6 +14,8 @@ end
 
 Rake::TestTask.new do |t|
   t.libs << "."
+  t.warning = true
+  t.verbose = true
   t.test_files = FileList['test/test*.rb']
 end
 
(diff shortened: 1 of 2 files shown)


iainbeeston:
Actually, we need https://github.com/sporkmonger/addressable/pull/180 to be merged before this is feasible (addressable throws up a lot of warnings)

iainbeeston:
This is now good to merge, as addressable has been fixed. Could someone in the core team please check it over? (TL;DR the test suite now shows ruby warnings if there are any)

RST-J:
:+1:

iainbeeston:
Thanks @RST-J - I think this is a fairly minor change so I'll merge it now

PR #234, opened 2015-02-13
Ignore fragments in schema caching. Fixes #233

### LINKED ISSUE #233: Absoluteness correction
`load_ref_schema` doesn't remove fragments, so loading schemas from absolute URIs in `"$ref"` values fails.

This issue opened just for reference, I have code for fixing it which I'll add via PR shortly.

(For reference this arose while trying to validate swagger v2 documents using their schema while having remote retrieval turned off)

jphastings:
If the URIs used in a `$ref` are absolute, but include fragments, then the fragment is not removed when the validator searches its cache to see what schemas are loaded, leading to duplication of schemas in the cache.

This code & test fixes that issue.

### DIFF
--- a/lib/json-schema/validator.rb
+++ b/lib/json-schema/validator.rb
@@ -129,7 +129,7 @@ module JSON
 
     def load_ref_schema(parent_schema, ref)
       schema_uri = absolutize_ref_uri(ref, parent_schema.uri)
-
+      schema_uri.fragment = ''
       return true if self.class.schema_loaded?(schema_uri)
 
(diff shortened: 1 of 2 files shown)


iainbeeston:
Good catch! But can we put the fix in absolutize_ref_uri, after ref_uri is assigned?

@RST-J Do you think we should roll this into the 2.5.1 release? (It gets my vote)

RST-J:
Yes, let's add this.

jphastings:
Cheers guys, glad to have helped!

PR #235, opened 2015-02-26
Expose clear_cache option

danieldraper:
This pull-request is to fix https://github.com/ruby-json-schema/json-schema/issues/225


iainbeeston:
Travis is failing, after this change (not sure why this would cause it to fail however)

iainbeeston:
If we can figure out why the build is failing it gets my :+1:

hoxworth:
Build is passing now, :+1:

iainbeeston:
Thanks @hoxworth . Merging...

PR #237, opened 2015-02-28
Removed test files from gemspec

iainbeeston:
Should reduce the size of the gem. The default for new gems (created by bundler) is not to include test files. See https://github.com/bundler/bundler/pull/3207


hoxworth:
:+1:

iainbeeston:
It's a minor change so I'll merge now.

PR #238, opened 2015-03-06
Allow boolean false as a default property.

chrisandreae:


iainbeeston:
Good catch. Could you please add a test to show that false works as a default now? (I assume you're using Json Schema to insert defaults?)

RST-J:
:+1:

hoxworth:
:+1:

iainbeeston:
:+1:

PR #240, opened 2015-03-06
Use old hash syntax in tests

iainbeeston:
Otherwise the ruby 1.8 travis build fails

### DIFF
--- a/test/support/strict_validation.rb
+++ b/test/support/strict_validation.rb
@@ -22,9 +22,9 @@ module StrictValidation
   end
 
   def test_strict_error_message
-    schema = { type: 'object', properties: { a: { type: 'string' } } }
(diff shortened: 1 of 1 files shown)


iainbeeston:
Incidentally, the ruby 1.8 build still fails even after this change (but at least it can be parsed). Merging #239 should help, because right now it's failing because on master we don't catch `Timeout::Error`s when loading remote schemas (this happens on 1.8 because in Ruby 1.9 a timeout is a runtime error, whereas in ruby 1.9+ it's a type of standard error, and in master we catch standard errors)

RST-J:
:+1:

iainbeeston:
I'll go ahead and merge this. Doesn't seem risky or controversial

PR #244, opened 2015-03-12
3.x rely on string keys during validation

RST-J:
This changes the validator to internally always rely on string keys in the given data.
It also adds the option to transform keys and values in passed in data into strings.


RST-J:
Moo, wrong target

Issue #243, opened 2015-03-11
v3: Handling of keys (strings, symbols, both)


RST-J:
To pick up on the discussion about hash keys that started in #161: How do we want to handle hash keys in the next major release?

I think, this should be changed soon because its a very central issue that affects nearly everything to come.

I suggest, that we internally always use string keys in every data structure (data and schemas).
If we decide to drop support for symbols, then this is it. But this will raise problems with insert defaults (#184) if we decide to keep support for that.

Otherwise we need to provide an option that indicates whether to convert all keys in the data to symbols after the validation process if it succeeded.
I'm in favour of the latter (if there are no big obstacles along the way), because its easy for the user and simplifies insert defaults (if still supported).

Quotes from @iainbeeston from #161
> 1. If we symbolize keys, we should also symbolize any other reference to properties in the schema (for example, required should have a symbol or array of symbols, so should depends).

My idea is to have a flag ``:symbolize_keys`` which defaults to ```false``` and if set to ```true``` converts the keys in the validated data to symbols **after** validation. During validation everything will be strings, so we do not need to convert anything in schemas. If we do not and use cached schemas, we'd have to convert the schemas used all the time or store them in a symbolized and a stringified version.

Thoughts?

iainbeeston:
I think having a :symbolize_keys option is a good idea.

If using strings internally causes problems for defaults I would prefer dropping support for defaults.

There is another option, which is that we don't convert symbols to strings at all. JSON has no concept of strings, and no Json parser that I'm aware of produces symbol keys, so I suspect that symbolized hashes are rarely used in practice. It's also easy (and less error probe) for a caller to stringify keys than it is for us to do it automatically. So what I'm suggesting is that we don't do the conversion at all and raise an error if we find a symbol key. This would be a breaking change, but then this is for the v3 release, right?

RST-J:
Yes. I actually think, this should be for v3 no matter what we decide to do because this will likely break current behaviour in some way.

I can also imagine that it would be easier to remove support for symbols (and also everything related to defaults) first, even if we want to have it back in in some way later, because the core validation code gets cleaned up.

RST-J:
Ok, after looking into this it seems that it isn't as huge as I expected it to be and more important we already used [string keys](https://github.com/ruby-json-schema/json-schema/blob/3.x/lib/json-schema/validator.rb#L580) during validation.

Actually, the tests will still pass with only the changes in the attribute classes.

[...]

PR #245, opened 2015-03-12
v3.x: Use string keys during validation. Add flag to stringify passed data

RST-J:
This changes the validator to internally always rely on string keys during validation and relates to #243.
It also adds a flag to convert keys and values in given data to strings.


RST-J:
Actually it doesn't introduce but changes how only string keys are used internally (see comment in #243).

iainbeeston:
I'm a bit confused. In this branch stringify no longer stringify ws the hash, but parses it. I can't see where it's actually stringified any more?

iainbeeston:
I like the changes, except I don't understand how stringify works if it just parses the data passed. Does that not mean that the input data will be converted to a Json schema object? And does it mean that we'll parse the same input hash multiple times?

RST-J:
The stringify method does not simply parse the data, it first converts it to json: ``JSON::Validator.parse(schema.to_json)``. As far as I know this does exactly one thing - convert symbol keys and values to string. If there is anything else that I wasn't aware of, I'd go back to the old implementation. Do you know about another case?

Stringify is not only related to schemas and was also used to stringify given in data before. I think it is in the wrong location, maybe it should go into some Util module but I had no idea where to place it exactly so I left it where it is for the moment.

As far as I can see we won't parse the same input multiple times. Stringify is used once at the beginning of schema initialization which makes sure that everything in schemas is converted to string keys/values and once to initialize data which also made sure, that every symbol key and value is a string during validation (this is what I meant by we already had that without the code in the attributes being aware of that).

iainbeeston:
Ah I see what stringify is doing now. However, taking a hash, serialising it to a string, then parsing it back again seems like a bad idea, performance-wise. Would it not be better to fix the existing stringify function to convert any symbol to a string?

Sorry, I was wrong about stringify being called twice. However, could we avoid calling it in the initializer? In master, the only place we modify the data is in initialize_data. This makes it easy to test and to reason about. It also keeps the initializer small, with no real work being done until validation begins. It's a point of style but I think it's worth keeping the separation that we currently have (if possible).

iainbeeston:
I've been thinking more about this (see #246) and now I think we probably should aim to remove inserting defaults in 3.0. It wouldn't be hard for users to insert defaults themselves using either [ActiveSupport's deep_merge](http://api.rubyonrails.org/classes/Hash.html#method-i-deep_merge) or maybe [Map](https://github.com/ahoward/map), and by letting them insert the defaults we avoid any ambiguity (plus we don't have to maintain that code and can focus on validation). Does that sound ok?

iainbeeston:
Oh, and I'd also suggest we stop stringifying keys, and just expect that the hash we're given is a json hash (ie. that only contains datatypes that are available in json). This would also simplify the codebase, and it's not hard for users to stringify for themselves (possibly by calling `to_json` then `parse` as you suggested).

In both cases we could probably add instructions to the readme to explain how to do it without json-schema.

RST-J:
I agree, its not hard to stringify for the users themselves. However, as we have this routine already, we could also keep it and make it available through ``JSON::Validator.stringify`` or similar. But it would be up to the user, to do that before validation, the validation itself wouldn't do that anymore.
The reason to keep our implementation is, that it also converts symbol values to strings.

iainbeeston:
You know, activesupport's as_json method is basically a superior version of
this. It's defined on a number of base classes (Array, Hash, Symbol etc)
and it recursively converts all symbols to strings, both symbol keys and
symbol values. What's more, many more complex classes already implement
as_json. I'd be tempted to say we should advise users to use that, as it
will always have better support than anything we can implement ourselves.

RST-J:
Ah, cool. No, I didn't know about this. But then I take back what I said and vote for the opposite ;).

iainbeeston:
I only realised today! (I've even written code before that's includes
symbols in the result of as_json! Whoops!)

I like the other changes in this PR though. I wonder if we could remove
stringification then update all tests to use strings instead of symbols
(might be a big job).

RST-J:
Should we drop the ``:parse_data`` option as well then? I think so. Especially when we also drop support for insertion of defaults we have no need to initialize given data at all.

iainbeeston:
Yes I agree, that makes sense

RST-J:
Done, I think. What to write into the README? Or do we want to rethink the README and start putting more detailed stuff into the wiki, as @hoxworth suggested before?

iainbeeston:
I think add something into the readme now. It doesn't have to be perfect
(we can revise the whole readme later), but at least we can't forget if we
add something now

TheMeier:
Closing due to inactivity. Please re-open if this is still relevant

PR #248, opened 2015-03-14
Codeclimate test coverage

iainbeeston:
This will start sending test coverage data to code climate, so you can see it inline, alongside code quality metrics.

It's set to merge into the 3.x branch, because codeclimate requires multijson on every build, making it impossible to test json-schema without multijson. But in 3.x we don't use multijson, so there's no side-effect.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,6 @@ matrix:
   include:
     - rvm: "2.2"
       gemfile: "gemfiles/Gemfile.uuidtools.x"
+env:
+  global:
+    - CODECLIMATE_REPO_TOKEN=f06667b968b44c85d8bacf7321b323574188f6771d0d3534e0cb9ddaade23dd7
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -1,5 +1,17 @@
+if ENV['CI'].to_s == 'true'
+  require "codeclimate-test-reporter"
+  CodeClimate::TestReporter.start
+end
+
 require 'minitest/autorun'
 require 'webmock/minitest'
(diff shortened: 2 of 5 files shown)


iainbeeston:
Here's what it looks like on [codeclimate](https://codeclimate.com/github/ruby-json-schema/json-schema)

screen shot 2015-03-14 at 13 56 41

RST-J:
:+1:

.travis.yml
```diff
@@ -17,3 +17,6 @@ matrix:
   include:
     - rvm: "2.2"
       gemfile: "gemfiles/Gemfile.uuidtools.x"
+env:
+  global:
+    - CODECLIMATE_REPO_TOKEN=f06667b968b44c85d8bacf7321b323574188f6771d0d3534e0cb9ddaade23dd7
```
mberlanda: Why this is not exported as an ENV var in Travis UI? It should be safer

test/test_helper.rb
```diff
@@ -1,6 +1,18 @@
+if ENV['CI'].to_s == 'true'
+  require "codeclimate-test-reporter"
+  CodeClimate::TestReporter.start
```
mberlanda: It looks like this implementation is deprecated:
https://github.com/codeclimate/ruby-test-reporter/blob/master/README.md

This could be done with SimpleCov:
https://docs.codeclimate.com/v1.0/docs/configuring-test-coverage

This change could be added to the `.travis.yml`

bastelfreak:
I'm closing this since we now have codecov reporting

PR #252, opened 2015-04-10
Explicitly notes :strict overrides any required properties set in schema

JKGisMe:
This tiny change in the readme just makes it clear that strict mode overrides the `required` properties set in the schema. The text said all properties were `required by default`, which I assumed meant that they would be required unless the schema explicitly specified otherwise. A very small but potentially frustrating distinction.

If others are like me this could save them some time debugging their complex schema & fragments.


iainbeeston:
Thanks for the correction!

PR #255, opened 2015-05-14
Speed up JSON::Validator.validate

mjc:
Addressible's URI parsing shows up at the top of the profiler for our use case on JRuby, as well as YAML dumping.

This patch
1. adds a cache to ```JSON::Util::URI.normalized_uri```
2. adds a cache for ```Addressable::URI.parse``` in ```JSON::Util::URI.parse```
3. prevents clearing an already cleared URI fragment in JSON::Schema#initialize
4. uses Hash#hash for calculating uuids instead of serializing the hash.

There are many further gains to be had still, especially if we can find a way to remove the mutex around initializing a schema, as that is proving to be terrible for performance on JRuby in a multi-threaded environment.

I'm pretty much finished with this pass - almost 8x is a pretty nice gain.

benchmark: https://github.com/mjc/json-schema-perf

MRI 2.2.2 results: **2.89x faster from file, 5.55x faster from hash.**

JRuby 1.7.20 results: **2.38x faster from file, 7.86x faster from hash.**

#### MRI 2.2.2 before
```
Calculating -------------------------------------
json-schema from file
                       181.000  i/100ms
json-schema from hash
                        87.000  i/100ms
-------------------------------------------------
json-schema from file
                          1.962k (± 6.3%) i/s -     58.644k
json-schema from hash
                        938.211  (± 6.2%) i/s -     28.101k

Comparison:
json-schema from file:     1961.8 i/s
json-schema from hash:      938.2 i/s - 2.09x slower
```

#### MRI 2.2.2 after
```
json-schema from file
                          5.668k (± 7.4%) i/s -    169.162k
json-schema from hash
                          5.210k (± 7.1%) i/s -    155.696k


Comparison:
json-schema from file:     5668.0 i/s
json-schema from hash:     5210.2 i/s - 1.09x slower
```
#### JRuby 1.7.20 before
```
Calculating -------------------------------------
json-schema from file
                       200.000  i/100ms
json-schema from hash
                       138.000  i/100ms
-------------------------------------------------
json-schema from file
                          2.075k (± 6.1%) i/s -     62.000k
json-schema from hash
                          1.401k (± 5.4%) i/s -     41.952k

Comparison:
json-schema from file:     2074.9 i/s
json-schema from hash:     1401.0 i/s - 1.48x slower
```

#### JRuby 1.7.20 after
```
Calculating -------------------------------------
json-schema from file
                       474.000  i/100ms
json-schema from hash
                         1.002k i/100ms
-------------------------------------------------
json-schema from file
                          4.942k (± 5.7%) i/s -    147.888k
json-schema from hash
                         10.773k (± 6.5%) i/s -    322.644k

Comparison:
json-schema from hash:    10773.1 i/s
json-schema from file:     4942.2 i/s - 2.18x slower
```

### DIFF
--- a/lib/json-schema/util/uri.rb
+++ b/lib/json-schema/util/uri.rb
@@ -4,10 +4,10 @@ module JSON
   module Util
     class URI
       include Singleton
-      attr_accessor :parse_cache, :normalized_cache
+      attr_accessor :cache
 
       def normalized_uri(uri)
-        uri = self.class.parse(uri) unless uri.is_a?(Addressable::URI)
+        uri = Addressable::URI.parse(uri) unless uri.is_a?(Addressable::URI)
         # Check for absolute path
         if uri.relative?
           data = uri.to_s
@@ -17,17 +17,12 @@ module JSON
         uri
       end
 
-      # These caches create a race condition in multithreaded environments
+      # This cache creates a race condition in multithreaded environments
       # This is OK because the worst case result of the race is
       # a cache miss
-      def self.parse(uri)
-        instance.parse_cache ||= {}
-        instance.parse_cache[uri.to_s.freeze] ||= Addressable::URI.parse(uri)
-      end
-
       def self.normalized_uri(uri)
-        instance.normalized_cache ||= {}
-        instance.normalized_cache[uri.to_s.freeze] ||= instance.normalized_uri(uri)
+        instance.cache ||= {}
+        instance.cache[uri] ||= instance.normalized_uri(uri)
       end
     end
   end
--- a/lib/json-schema/util/uri.rb
+++ b/lib/json-schema/util/uri.rb
@@ -17,6 +17,9 @@ module JSON
         uri
       end
 
+      # This cache creates a race condition in multithreaded environments
+      # This is OK because the worst case result of the race is
+      # a cache miss
       def self.normalized_uri(uri)
         instance.cache ||= {}
         instance.cache[uri] ||= instance.normalized_uri(uri)
--- a/lib/json-schema/util/uri.rb
+++ b/lib/json-schema/util/uri.rb
@@ -11,7 +11,7 @@ module JSON
         # Check for absolute path
         if uri.relative?
           data = uri.to_s
-          data = "#{Dir.pwd}/#{data}" if data[0,1] != '/'
+          data = "#{Dir.pwd}/#{data}" if data[0, 1] != '/'
           uri = Addressable::URI.convert_path(data)
         end
         uri
@@ -22,7 +22,7 @@ module JSON
       # a cache miss
       def self.normalized_uri(uri)
         instance.normalize_cache ||= {}
-        instance.normalize_cache[uri] ||= instance.normalized_uri(uri)
+        instance.normalize_cache[uri.to_s] ||= instance.normalized_uri(uri)
       end
 
       def self.parse(uri)
--- a/lib/json-schema/util/uri.rb
+++ b/lib/json-schema/util/uri.rb
@@ -4,7 +4,7 @@ module JSON
   module Util
     class URI
       include Singleton
-      attr_accessor :cache
+      attr_accessor :normalize_cache, :parse_cache
 
       def normalized_uri(uri)
         uri = Addressable::URI.parse(uri) unless uri.is_a?(Addressable::URI)
@@ -21,8 +21,13 @@ module JSON
       # This is OK because the worst case result of the race is
       # a cache miss
       def self.normalized_uri(uri)
-        instance.cache ||= {}
-        instance.cache[uri] ||= instance.normalized_uri(uri)
+        instance.normalize_cache ||= {}
+        instance.normalize_cache[uri] ||= instance.normalized_uri(uri)
+      end
+
+      def self.parse(uri)
+        instance.parse_cache ||= {}
+        instance.parse_cache[uri.to_s] ||= Addressable::URI.parse(uri)
       end
     end
   end
--- a/lib/json-schema/util/uri.rb
+++ b/lib/json-schema/util/uri.rb
@@ -1,7 +1,12 @@
+require 'singleton'
+
 module JSON
   module Util
-    module URI
-      def self.normalized_uri(uri)
+    class URI
+      include Singleton
+      attr_accessor :cache
+
+      def normalized_uri(uri)
         uri = Addressable::URI.parse(uri) unless uri.is_a?(Addressable::URI)
         # Check for absolute path
         if uri.relative?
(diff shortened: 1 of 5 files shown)


mjc:
I'll update the pull request description shortly with new benchmark numbers, found a bug that was an even more dramatic speedup.

mjc:
I also need to research finding ways to test concurrency and remove the mutex around schema initialization safely.

lib/json-schema/util/uri.rb
```diff
@@ -11,6 +16,11 @@ def self.normalized_uri(uri)
         end
         uri
       end
+
+      def self.normalized_uri(uri)
+        instance.cache ||= {}
+        instance.cache[uri] ||= instance.normalized_uri(uri)
```
iainbeeston: The caching here is totally unsafe in a multithreaded environment. You're going to end up with race conditions. But the side effect of that is (I admit) fairly minor here because the caller will always get the correct result.

However, I've had a long think about whether caching is the right approach (since you raised this pull request) and I think you're probably right to add it.

lib/json-schema/validator.rb
```diff
@@ -474,6 +474,8 @@ def merge_missing_values(source, destination)
 
         if @@json_backend == 'yajl'
           @@serializer = lambda{|o| Yajl::Encoder.encode(o) }
+        elsif @@json_backend == 'json'
+          @@serializer = lambda{|o| JSON.dump(o) }
```
iainbeeston: Why change the serialisation method? So far as I'm aware (as I could be wrong) it's only used to generate a cache key for a given schema. I assume its performance related? If so I wonder if we could change to using Hash#hash rather than serializing the whole schema out to a string and generating the key from that.

iainbeeston:
With regards to the mutex, I'm not even sure what it's purpose is. So far as I can tell it ensures that only one validator can be initialised at once. But it's not protecting access to any class variables, and it's not protecting a validator from being used by several threads concurrently and getting into an invalid state.

It's possible the answer is in the git history, though I'm not in front of a computer right now.

@hoxworth Do you remember why the mutex was introduced?

lib/json-schema/util/uri.rb
mjc: @iainbeeston right, the only effect of the race condition is a cache miss, and that's OK.

lib/json-schema/validator.rb
mjc: @iainbeeston it was the vast majority of the profile.

mjc: I changed the methods that were calling serialize and passing the serialized result to fake_uuid (lines 542 and 549) to use schema.hash.to_s, and this was the result:

#### MRI 2.2.2 (within error range)
```
json-schema from file
                          5.668k (± 7.4%) i/s -    169.162k
json-schema from hash
                          5.210k (± 7.1%) i/s -    155.696k


Comparison:
json-schema from file:     5668.0 i/s
json-schema from hash:     5210.2 i/s - 1.09x slower
```

#### JRuby 1.7.20 (1.21x faster for hash, slightly faster for file)
```
Calculating -------------------------------------
json-schema from file
                       420.000  i/100ms
json-schema from hash
                       830.000  i/100ms
-------------------------------------------------
json-schema from file
                          4.339k (± 3.4%) i/s -    130.200k
json-schema from hash
                          8.414k (± 3.1%) i/s -    252.320k

Comparison:
json-schema from hash:     8414.0 i/s
json-schema from file:     4339.4 i/s - 1.94x slower
```

mjc:
I'm pretty much done with this pass if we can't remove the mutex.

RST-J:
To me that looks OK and I guess using some more memory is a good trade for increased performance.

lib/json-schema/util/uri.rb
```diff
@@ -11,6 +16,19 @@ def self.normalized_uri(uri)
         end
         uri
       end
+
+      # This cache creates a race condition in multithreaded environments
+      # This is OK because the worst case result of the race is
+      # a cache miss
+      def self.normalized_uri(uri)
+        instance.normalize_cache ||= {}
+        instance.normalize_cache[uri] ||= instance.normalized_uri(uri)
```
iainbeeston: One last question - should the cache key here be "uri.to_s"? (As it is for the parse cache?)

lib/json-schema/util/uri.rb
```diff
@@ -1,7 +1,12 @@
+require 'singleton'
+
 module JSON
   module Util
-    module URI
-      def self.normalized_uri(uri)
+    class URI
+      include Singleton
+      attr_accessor :normalize_cache, :parse_cache
```
iainbeeston: Also, it might be good to make these accessors "protected" methods (they're globally available caches, so it would be good to prevent access from outside this class)

iainbeeston:
I've added a couple more questions, but overall this looks great and I'll be happy to merge once I hear back from @mjc. I especially appreciate the benchmark times that accompany the changes (thanks @mjc!)

lib/json-schema/util/uri.rb
mjc: nice catch!

lib/json-schema/util/uri.rb
mjc: will do.

mjc:
@iainbeeston no problem! Happy to help. Trying to improve performance if you don't know where the bottlenecks are and whether you made a useful difference is silly.

mjc:
Weird, I think maybe the .parse cache just has to go. :(

iainbeeston:
So it works when the key is a uri object but not when it is a string?

mjc:
It looks like the objects get modified in incompatible ways sometimes? The
error on Travis is proving hard to track down

mjc:
Do you think it'd be possible to use the stdlib's URI stuff?
http://ruby-doc.org/stdlib-2.2.2/libdoc/uri/rdoc/URI.html

iainbeeston:
Is it possible to freeze the URI object? Would that help?

The gem used to do that but although it's faster it has trouble with
unusual URIs (file URIs I think are an example of that, but generally
addressable is more capable)

lib/json-schema/util/uri.rb
RST-J: And can't you use the class directly instead of a singleton or switch back to the module and use private cache constants?

iainbeeston:
@RST-J and @mjc Could you please take a look at #285 , which is my attempt to salvage this and make it releasable?

iainbeeston:
I'm closing this for now. I've merged most of the changes with the exception of using has to generate uuids (which gave a good speed improvement but feels risky to me)

PR #256, opened 2015-05-15
allow OpenStruct to be validated as an object

JonnieCache:
Fairly self explanatory.

Sorry there aren't any tests, I tried to write some but the metaprogramming in `test/support/type_validation.rb` got in my way, don't have time to mess around with it atm, sorry :(

### DIFF
--- a/lib/json-schema/attribute.rb
+++ b/lib/json-schema/attribute.rb
@@ -29,10 +29,10 @@ module JSON
         "number" => Numeric,
         "integer" => Integer,
         "boolean" => [TrueClass, FalseClass],
-        "object" => Hash,
+        "object" => [Hash, OpenStruct],
         "array" => Array,
         "null" => NilClass,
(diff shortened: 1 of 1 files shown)


JonnieCache:
just spotted my hideous error, fixing it now...

RST-J:
I'm not really in favor of supporting another type as object. We recently decided that we want to keep the library clean and slim (which is why e.g. symbol support will be dropped in the next major release) and this would be a step in the opposite direction.
Couldn't you just write a wrapper around json-schema which goes through your objects to be validated and replaces every struct with ``struct.to_h.to_json``?

iainbeeston:
Yes, I agree with @RST-J - I'd recommend ensuring that your hash is valid JSON (by using #to_json) before validating

JonnieCache:
Fair enough, that works for me.

PR #263, opened 2015-09-08
Added type casting for limiting min/max

elliott-davis:
Both 0 and "0" are valid json integers when representing min/max.
I added a cast to the limit to ensure proper comparison.

### DIFF
--- a/lib/json-schema/attributes/limit.rb
+++ b/lib/json-schema/attributes/limit.rb
@@ -15,7 +15,7 @@ module JSON
 
       def self.invalid?(schema, data)
         exclusive = exclusive?(schema)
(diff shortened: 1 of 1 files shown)


RST-J:
Could you please add a reference which backs your statement?
To me it seems like you'd like to have it this way, but the [spec](http://json-schema.org/latest/json-schema-validation.html#anchor13) clearly says it has to be a JSON number.

elliott-davis:
I'm sorry - I think I pinned the blame on json-schema when it really belongs with raml-mocker.  I'm going to close this as the spec is pretty clear it needs to be an integer.

PR #267, opened 2015-10-02
Create method to Validator that accepts data argument.

glennpratt:
Currently the Validator class requires data be passed into the constructor, though it doesn't do anything important with it then.

In my case, I'm validating in a server and I see no reason to parse the schema over and over for different inputs, especially since it uses two mutexes in the initialize method.

Let me know if I missed something and Validator simply must accept data at construction.

Is `#validate_data(data)` the way to go, or should I change the signature of `#validate` to take an optional argument? The former seemed like a clean change with little API change.

### DIFF
--- a/lib/json-schema/validator.rb
+++ b/lib/json-schema/validator.rb
@@ -15,6 +15,20 @@ module JSON
 
   class Validator
 
+    class Processor
+      def initialize
+        @errors = []
+      end
+
+      def validation_error(error)
+        @errors.push(error)
+      end
+
+      def validation_errors
+        @errors
+      end
+    end
+
     @@schemas = {}
     @@cache_schemas = false
     @@default_opts = {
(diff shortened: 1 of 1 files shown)


glennpratt:
I wrote this little script to look for any obvious threading issues.

https://gist.github.com/glennpratt/bf254a896c9b0f15b812

RST-J:
To me this looks fine. :+1:

glennpratt:
This needs work, have issues with shared @errors instance var in testing.

glennpratt:
OK, pushed a commit that should fix that issue.

Need more tests around that apparently.

glennpratt:
Note: This PR is sparse to help decide if this is a good direction to move it.

If it is, I'll update usage, documentation and tests.

So, do people think this is the right direction?

(Thanks @RST-J)

jpmckinney:
Yes, this is the right direction. Re-initializing a schema everytime you validate a record is extremely slow. The schema should be initialized once, and then used to validate any number of records.

iainbeeston:
Yes I agree with the direction completely. I would actually change the signature of the validate method. We already have a 3.x branch for exactly this kind of backwards incompatible change.

Once you're happy with the approach could open a new pull request on the 3.x branch?

TheMeier:
activity timeout

PR #268, opened 2015-10-18
Added new option :restrict_additional_properties. Datas can be a subs…

tak2siva:
…et of schema which is not possible with :strict

### DIFF
--- a/lib/json-schema/attributes/properties.rb
+++ b/lib/json-schema/attributes/properties.rb
@@ -33,6 +33,21 @@ module JSON
           end
         end
 
+        if options[:restrict_additional_properties]
+          diff = data.select do |k, v|
+            k = k.to_s
+            !schema['properties'].has_key?(k)
+          end
+
+          if diff.size > 0
+            properties = diff.keys.join(', ')
+            message = "The property '#{build_fragment(fragments)}' contained undefined properties: '#{properties}'"
+            validation_error(processor, message, fragments, current_schema, self, options[:record_errors])
+          end
+
+          return
+        end
+
         # When strict is true, ensure no undefined properties exist in the data
         return unless options[:strict] == true && !schema.key?('additionalProperties')
 
--- a/lib/json-schema/validator.rb
+++ b/lib/json-schema/validator.rb
@@ -26,7 +26,8 @@ module JSON
       :insert_defaults => false,
       :clear_cache => true,
       :strict => false,
-      :parse_data => true
+      :parse_data => true,
+      :restrict_additional_properties => false
     }
     @@validators = {}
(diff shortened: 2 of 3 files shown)


tak2siva:
Our application contracts are mostly a subset of the original. Looks like there is no way in json-schema to validate our needs(up to my knowledge). So added the option :restrict_additional_properties which will only check for additional properties. 

    schema = {
      "$schema" => "http://json-schema.org/draft-04/schema#",
      "properties" => {
        "a" => {"type" => "string"},
        "b" => {"type" => "string"}
      }
    }

    data = {"a" => "a"}
    assert(JSON::Validator.validate(schema,data,:restrict_additional_properties => true)) ## Success

    data = {"a" => "a", "d" => "d"}
    assert(JSON::Validator.validate(schema,data,:restrict_additional_properties => true)) ## Fail

RST-J:
If you simply want to make sure that there are no additional properties you should modify your schemas such that every object description has ``additionalProperties: false``.
As I understood your requirement, this should exactly do what you are looking for.

jlblcc:
This option would be useful when *developing* schemas and for testing. Currently, I use an alternative [validator ](https://github.com/geraintluff/tv4#the-banunknownproperties-flag) to perform this type of testing. Strict mode doesn't help with testing [partial examples](https://github.com/adiwg/mdJson-schemas/blob/master/examples/keywords.json), since *all properties are treated as required*.

tak2siva:
@RST-J  Nope. That didnt work.

    data = {"a" => "a", "b" => "b", "c" => "c"}
    assert(!JSON::Validator.validate(schema,data,:additionalProperties => false))  ## This test will fail 

    data = {"a" => "a", "b" => "b", "c" => "c"}
    assert(!JSON::Validator.validate(schema,data,:restrict_additional_properties => true))  ## This test will pass

RST-J:
No, I don't mean as an option to json-schema but as part of the schema you pass in:
```json
schema = {
  "$schema" => "http://json-schema.org/draft-04/schema#",
  "properties" => {
    "a" => {"type" => "string"},
    "b" => {"type" => "string"}
  },
  "additionalProperties" => false
}
```

tak2siva:
@RST-J  Isn't that tedious and repetitive doing it for all endpoint's ? Especially when an established big projects starts using this gem. Why not a simple config for that ?

iainbeeston:
Sorry @tak2siva but weak ready have far too many custom options to maintain.

If you find it repetitive, why not add additionalProperties: false programmatically (you can provide the schema as a ruby hash, as well as a string)

tak2siva:
No issues. Switched to an alternative long back.

PR #274, opened 2015-11-19
Developers often use Pathname objects for paths

krainboltgreene:
Strings are naive path names, and in Ruby we have an object specifically for path names called Pathname.

### DIFF
--- a/lib/json-schema/validator.rb
+++ b/lib/json-schema/validator.rb
@@ -512,7 +512,7 @@ module JSON
     end
 
     def initialize_schema(schema)
-      if schema.is_a?(String)
+      if schema.is_a?(String) || schema.is_a?(Pathname)
         begin
           # Build a fake URI for this
           schema_uri = Addressable::URI.parse(fake_uuid(schema))


iainbeeston:
I think really we should be moving away from explicit type checking and towards duck typing.

Looking at how [OpenURI](https://github.com/ruby/ruby/blob/trunk/lib/open-uri.rb#L29) handles this problem, it checks for the `open()` method (which is supported by `Pathname` and `IO` objects) and failing that it checks for `to_str()` (which is supported by strings and `Pathname` in ruby prior to 1.9).

This method is a bit of a mess, it should probably be checking for supported methods (like `OpenURI` does) rather than using rescue instead of conditionals and rescue statements for flow control.

RST-J:
I agree with @iainbeeston, I think it's the most flexible way and having rescue for control flow is really a code smell.

iainbeeston:
I'm going to close this for now. Feel free to open another PR

krainboltgreene:
This is rather disappointing.

iainbeeston:
If you have time to make a new pull request that uses duck-typing rather than explicit type checking and we'll be more than happy to look at it

PR #284, opened 2015-12-23
Added the orderedhash gem for ruby 1.8 builds

iainbeeston:
This should allow the 1.8 build to pass reliably as it makes hash keys
ordered, and many of the tests rely on hash keys appearing in creation
order.


iainbeeston:
This idea didn't work so well. Closing

PR #285, opened 2015-12-23
Redux: speed up JSON::Validator.validate

iainbeeston:
This is a rebase and update of #255 

I've pulled out the change to the way uuids are calculated (in #255 the uuid of a Hash is calculated by taking a SHA of `Hash.hash`, whereas in the original the SHA of the string of the hash is used) because I feel that the risk is high and a better solution would be to use a faster hashing algorithm.

However, I've expanded the use of caching to get around the speed problems in Addressable.

Worst case this should be faster than the current implementation.

### DIFF
--- a/lib/json-schema/attributes/ref.rb
+++ b/lib/json-schema/attributes/ref.rb
@@ -52,7 +52,7 @@ module JSON
           fragment_path = ''
           fragments.each do |fragment|
             if fragment && fragment != ''
-              fragment = Addressable::URI.unescape(fragment.gsub('~0', '~').gsub('~1', '/'))
+              fragment = JSON::Util::URI.unescaped_uri(fragment.gsub('~0', '~').gsub('~1', '/'))
               if target_schema.is_a?(Array)
                 target_schema = target_schema[fragment.to_i]
               else
--- a/lib/json-schema/attributes/ref.rb
+++ b/lib/json-schema/attributes/ref.rb
@@ -1,5 +1,6 @@
 require 'json-schema/attribute'
 require 'json-schema/errors/schema_error'
+require 'json-schema/util/uri'
 
 module JSON
   class Schema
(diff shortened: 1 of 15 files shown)


iainbeeston:
If I can get this reviewed I'd like to release as v2.6.0

lib/json-schema/attributes/ref.rb
```diff
@@ -21,21 +22,23 @@ def self.validate(current_schema, data, fragments, processor, validator, options
...
+            path = Pathname.new(path).cleanpath.to_s
+            temp_uri.merge!(:path => path, :fragment => fragment)
           else
             temp_uri = current_schema.uri.join(path)
+            temp_uri.fragment = fragment
           end
-          temp_uri.fragment = s['$ref'].split("#")[1]
         end
-        temp_uri.fragment = "" if temp_uri.fragment.nil?
+        temp_uri = temp_uri.merge(:fragment => "") if temp_uri.fragment.nil? || temp_uri.fragment.empty?
```
jpmckinney: If `temp_uri.fragment.empty?` is true, then will `temp_uri.merge(fragment: "")` change anything?

Also, can do `temp_uri.merge!(fragment: "")` instead of `temp_uri = temp_uri.merge(fragment: "")` to be consistent with preceding code.

lib/json-schema/util/uri.rb
```diff
@@ -3,15 +3,44 @@ module Util
...
+        end
+
+        normalized_uri
+      end
+
+      def self.parse(uri)
+        return uri if uri.is_a?(Addressable::URI)
+
+        @parse_cache ||= {}
+        parsed_uri = @parse_cache[uri]
```
jpmckinney: Wouldn't all this be more succinctly expressed as:

```ruby
@parsed_cache[uri] ||= Addressable::URI.parse(uri).freeze
```

lib/json-schema/attributes/ref.rb
iainbeeston: Let me answer those in turn:

1. If temp_uri.fragment is "", then the URI will have a trailing "#" symbol. If temp_uri.fragment is nil or empty then no "#" is appended. So doing this does change the behaviour
2. We can't use merge! here, because is temp_uri.relative? is false, then temp_uri has come directly from the cache in JSON::Util:URI, and uris stored in the cache are frozen. If you call merge! on a frozen uri it will raise an exception because you are modifying a frozen object. By using merge it will work with both branches of the code

lib/json-schema/util/uri.rb
iainbeeston: Good catch, I've corrected that now

iainbeeston:
I've refactored slightly to duplicate cached URIs rather than freezing them

iainbeeston:
(Because in addressable dup is cheaper than freeze)

lib/json-schema/schema/reader.rb
```diff
@@ -71,8 +70,6 @@ def read(location)
         JSON::Schema.new(JSON::Validator.parse(body), uri)
       end
 
-      # @param uri [Addressable::URI]
-      # @return [Boolean]
```
jpmckinney: Why delete the docstrings?

lib/json-schema/attributes/ref.rb
```diff
@@ -21,21 +22,23 @@ def self.validate(current_schema, data, fragments, processor, validator, options
...
-          elsif path[0,1] == "/"
-            temp_uri.path = Pathname.new(path).cleanpath.to_s
-          else
-            temp_uri = current_schema.uri.join(path)
+        temp_uri = JSON::Util::URI.parse(s['$ref'])
+        temp_uri.defer_validation do
+          if temp_uri.relative?
+            temp_uri.merge!(current_schema.uri)
+            # Check for absolute path
+            path, fragment = *s['$ref'].split("#")
```
jpmckinney: I don't think the `*` is necessary.

lib/json-schema/schema/reader.rb
iainbeeston: Because in 6967d99 I've tried to abstract away the fact that we use `Addressable::URI`, however, I'm probably kidding myself. So I've restored this until I have time to make a more water-tight solution.

lib/json-schema/attributes/ref.rb
iainbeeston: Interesting - I always thought that was necessary for array restructuring (but apparently only if one of the destination variables should be an array)

RST-J:
Looks fine to me :+1:

iainbeeston:
Thanks. I'll merge this and make a release

PR #286, opened 2015-12-25
Remove ruby warnings

teoljungberg:
For the unused varaiables `schema`, `data`, and `validator`

### DIFF
--- a/test/test_load_ref_schema.rb
+++ b/test/test_load_ref_schema.rb
@@ -18,23 +18,21 @@ class LoadRefSchemaTests < Minitest::Test
 
   def test_cached_schema
     schema_url = "http://example.com/schema#"
-    schema = {
-      "$ref" => schema_url
-    }
+    schema = { "$ref" => schema_url }
     data = {}
     load_other_schema
-    validator = JSON::Validator.new(schema, data)
+    _validator = JSON::Validator.new(schema, data)
+
     assert JSON::Validator.schema_loaded?(schema_url)
   end
 
   def test_cached_schema_with_fragment
     schema_url = "http://example.com/schema#"
-    schema = {
-      "$ref" => "#{schema_url}/properties/title"
-    }
+    schema = { "$ref" => "#{schema_url}/properties/title" }
     data = {}
     load_other_schema
-    validator = JSON::Validator.new(schema, data)
+    _validator = JSON::Validator.new(schema, data)
+
     assert JSON::Validator.schema_loaded?(schema_url)
   end
 end


iainbeeston:
Hey @teoljungberg 

This test is such a mess, isn't it? I'm not sure the original really explained the intent.

test/test_load_ref_schema.rb
```diff
@@ -12,29 +12,19 @@ def load_other_schema
           }
         }
       },
-      Addressable::URI.parse("http://example.com/schema#")
+      Addressable::URI.parse(url)
```
iainbeeston: This line can be removed - we parse the address then throw it away.

test/test_load_ref_schema.rb
```diff
@@ -12,29 +12,19 @@ def load_other_schema
...
   def test_cached_schema
     schema_url = "http://example.com/schema#"
-    schema = {
-      "$ref" => schema_url
-    }
-    data = {}
-    load_other_schema
-    validator = JSON::Validator.new(schema, data)
+    load_other_schema(schema_url)
     assert JSON::Validator.schema_loaded?(schema_url)
```
iainbeeston: I believe the original intent here was to:

1. Add `http://json-schema.org/draft-04/schema#` to the `JSON::Validator::schemas` cache
2. Call `JSON::Validator.new(schema, data)` with a new schema that refers to `http://example.com/schema#` in a `$ref`
3. Check that creating a new validator adds the referenced schema to `JSON::Validator::schemas`

test/test_load_ref_schema.rb
```diff
@@ -12,29 +12,19 @@ def load_other_schema
...
   def test_cached_schema_with_fragment
     schema_url = "http://example.com/schema#"
-    schema = {
-      "$ref" => "#{schema_url}/properties/title"
-    }
-    data = {}
-    load_other_schema
-    validator = JSON::Validator.new(schema, data)
+    load_other_schema(schema_url)
     assert JSON::Validator.schema_loaded?(schema_url)
```
iainbeeston: I believe the original intent here was to:

1. Add `http://json-schema.org/draft-04/schema#` to the `JSON::Validator::schemas` cache
2. Call `JSON::Validator.new(schema, data)` with a new schema that refers to `http://example.com/schema#/properties/title` in a $ref
3. Check that creating a new validator adds the parent schema (`http://example.com/schema#`) and not the specific property to `JSON::Validator::schemas`

iainbeeston:
I've tried to explain the intent inline.

I think your changes make both tests identical? And also I think it has lost the original intent of the tests

test/test_load_ref_schema.rb
teoljungberg: It doesn't seem like it. `JSON::Schema` expects 2 arguments, one is the schema as a hash, the second is the `schema_url` as a `URI`-like object.

teoljungberg:
@iainbeeston You are absolutely right.

In light of that, I think the only change needed here is to mark `validator` as an unused variable. I've updated the PR to reflect that. Am I correct to assume so?

test/test_load_ref_schema.rb
iainbeeston: Ah, my bad - I misread that line as being the return value

iainbeeston:
@teoljungberg Yes that's right. I'd be more than happy to merge that. I can't see the updated code though?

teoljungberg:
@iainbeeston I took some time away from the computer during the holidays, sorry I haven't responded.

I've updated the PR now with the recent changes after our discussion. I've also squashed the commits.

RST-J:
:+1:

PR #290, opened 2016-01-08
Tidy up tests

iainbeeston:
I've tidied up a few things in the test directory that were annoying me:

1. Made test classes consistently named `FooTest` (not `TestFoo`, `FooTests` or just `Foo`)
2. Renamed test files to match the test class that they contain
3. Moved `test_helper.rb` out of the `/test` directory and into `/test/support`
4. Renamed any tests that had inaccurate, vague or misleading names


iainbeeston:
Any thoughts on this? (My goal is to make it easier to know where to add new tests, by making the naming of existing ones more consistent)

iainbeeston:
@RST-J do you think this is a good idea? Or can you suggest any changes?

RST-J:
I'll have a look some time soon. I think this needs some time to think about. In general I like your idea and aim very much.

test/files_test.rb
```diff
@@ -11,18 +11,28 @@ def test_schema_from_file
     refute_valid schema_fixture_path('good_schema_1.json'), { "a" => "bad" }
   end
 
-  def test_data_from_file
+  def test_data_from_file_v3
```
RST-J: Shouldn't this preferably called `data_from_file_v3_test` according to the new naming scheme?

RST-J:
Not sooo soon, but finally. I think it's fine. :+1:

test/files_test.rb
iainbeeston: Test unit only runs methods starting with `test_` - so this is the correct method name. I've also checked the other test classes and they all seem to have correct test method names

PR #293, opened 2016-01-14
Made sure we include the module name for SchemaParseErrors

### LINKED ISSUE #292: NameError: uninitialized constant JSON::Validator::SchemaParseError
When one passes an object, instead of a valid schema to the library it throws
```
NameError:
       uninitialized constant JSON::Validator::SchemaParseError
```

This seems to be because `SchemaParseError` is under `JSON::Schema`, while the call site is in `JSON::Validator`.

See `lib/json-schema/validator.rb`, line 572 for the call site.

iainbeeston:
If you pass an unrecognized object for the schema, we should be raising a SchemaParseError, however instead right now a NameError is raised, beacuse we're not correctly qualifying the error class with it's module.

I've corrected the class/module name and added tests.

This fixes #292

### DIFF
--- a/lib/json-schema/validator.rb
+++ b/lib/json-schema/validator.rb
@@ -569,7 +569,7 @@ module JSON
         end
         Validator.add_schema(schema)
       else
-        raise SchemaParseError, "Invalid schema - must be either a string or a hash"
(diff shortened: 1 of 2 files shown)


RST-J:
:+1:

iainbeeston:
@davidnquach Sorry for the delay, but I've released this fix now as 2.6.1

Issue #265, opened 2015-09-15
Passing in a fragment with an even number of tokens can lead to undefined method `validate' for #<Hash...>


ckundo:
We're seeing an exception when validating a json response from a Rails controller. It seems to fail on every other token in the JSON fragment path:

``` ruby
schema_options = {
  fragment: "#/schemas/v1.reports.index.response/purchases",
  strict: true
}
data = JSON.parse(response.body)

JSON::Validator.fully_validate(
  "schema.json",
  data,
  schema_options,
)
```

Raises:

```
# => ...undefined method `validate' for #<Hash...>
```

But for example, this fragment does not raise the exception:

```
"#/schemas/v1.reports.index.response/purchases/items"
```

We have monkey-patched this as follows:

``` ruby
module JSON
  class Validator
    def schema_from_fragment(base_schema, fragment)
      # ...

      fragments.each do |f|
        if base_schema.is_a?(JSON::Schema) #test if fragment is a JSON:Schema instance
          if !base_schema.schema.has_key?(f)
            raise JSON::Schema::SchemaError.new("Invalid fragment resolution for :fragment option")
          end

          # HERE:
          if base_schema.schema[f].is_a? Hash
            base_schema = JSON::Schema.new(base_schema.schema[f], schema_uri, @options[:version])
          else
            base_schema = base_schema.schema[f]
          end
        elsif base_schema.is_a?(Hash)
          # ...
```

That is, this line https://github.com/ruby-json-schema/json-schema/blob/master/lib/json-schema/validator.rb#L91 should always create a new `JSON::Schema`.

rickreilly:
+1

RST-J:
Would you please provide a test case, example schema and data which fails?

jrgns:
I'm seeing the same issue, but it's not due to the number of tokens. It happens when the second last token points to a proper schema, and the last token points to a Hash.

The bug exists because the resulting base_schema get's converted into a JSON::Schema in all the if clauses except the first one: https://github.com/jrgns/json-schema/blob/master/lib/json-schema/validator.rb#L89

jrgns:
This might be by design, though, since you can use the fragment to point to something that is not a schema like a property or a required array.

The reason I ran across it is that I'm trying to validate both the input and the output of an API call using the [hyper-schema](http://json-schema.org/draft-04/hyper-schema):

```
JSON::Validator.validate!('schema.json', input, strict: true, fragment: '#/definitions/users/links/0/schema')
[...]

PR #217, opened 2015-01-16
Fix validator when the specified fragment includes array

take:

PR #294, opened 2016-01-16
fix: Ensure the schema is a JSON::Schema

jrgns:
This is related to #265 . It doesn't have tests, but it shows a quick fix that should work.


jrgns:
Actually, #217 is a better fix for this, and includes tests.

Issue #103, opened 2014-07-11
Incorrect behavior of strict vs. default modes


pdlug:
The behavior of strict mode is a bit misleading (and incorrect IMHO). JSON schema provides `required` already but in the default execution of the json-schema validator it ignores this property. The expected behavior is that default validation checks that any keys present in the document being validated match the definition in the schema and that all keys listed as `required` in the schema are present. Strict mode would then make a bit more sense to change all keys to required though I'd argue that's a bit misleading as well and strict mode should really just imply that there are no keys present in the validation target that are not defined in the schema as well. Specifying that all keys present in the schema are required really should be a separate option since this is a different validation case (perhaps `require_all: true`?).

In summary I propose:

Default validation:
- Invalid unless all keys in the input document match the definition in the schema.
- Invalid unless all keys listed as required in the schema are present. 
- Keys present in the input document but not present in the schema are ignored (i.e don't affect validation).

Strict validation:
- Default rules + invalid if any keys are present in the input document but not defined in the schema.

Require all:
- Strict validation + invalid unless all keys defined in the schema are present in the input document.

I'm happy to take a stab at a patch if I can get some consensus that this is the right thing to do.

tristil:
FWIW there's a [thoughtbot post]( https://robots.thoughtbot.com/validating-json-schemas-with-an-rspec-matcher) where they use `strict: true` so if people copy that matcher they may lose some time trying to figure out why for example the parser isn't respecting `required` lists, etc. I don't really get what strict is for either, it ends up just distorting the schema you've specified.

freiden:
Sorry for being late to the party but is strict option is still needed since behaviors are defined in schemas?

RST-J:
@freiden What exactly do you mean? Strict is a makro so to say which prohibits any properties not covered by the schema (default in JSON schema is to accept any properties about which nothing is stated in a schema) and at the same time make every property required which is listed in a schema (properties not explicitly stated as required can be missing by default).

It is not necessary to have this as one can achieve this by hand. However, it was introduced as shortcut so one has not to do this tedious steps to convert relaxed schemas to strict ones, iirc.

freiden:
If I understand well the purpose is to validate schemas based on the specfication, so validations of fields should be managed by the schema to prevent unwanted behavior in context where you doesn't use this gem for validations. 
For example, I'm using the gem to validate my schema during tests but in my API I don't need the validation when sending my data the rest of the time.

cec:
[...]

PR #295, opened 2016-01-21
add :require_all option in addition to :strict

pkarman:
We want to test if example JSON adheres to the schema, esp for required and additionalProperties. However, the current `:strict` behavior is too strict.

The current behavior of `:strict` turns on an implied required for all properties. This PR makes that implicit requirement explicit, and off by default.

To get the current behavior, this PR would require:
```ruby
JSON::Validator.validate(schema, {"a" => 1}, :strict => true, :require_all => true)
```
instead of simply `:strict => true`.

### DIFF
--- a/README.textile
+++ b/README.textile
@@ -82,7 +82,9 @@ JSON::Validator.validate('user.json', data, :list => true)
 
 h3. Strictly validate an object's properties
 
-With the <code>:strict</code> option, validation fails when an object contains properties that are not defined in the schema's property list or doesn't match the <code>additionalProperties</code> property. Furthermore, all properties are treated as <code>required</code> regardless of <code>required</code> properties set in the schema.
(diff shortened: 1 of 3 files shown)


pkarman:
fails too many tests this way. will re-open with reverse logic (keeping original behavior by default).

cec:
@pkarman I second your opinion that we need a proper `strict` mode. 
I posted again on #103 where this was extensively discussed and also referenced your PR.

pkarman:
@cec thanks. I hadn't realized there was such a history here. I implemented this PR in complete ignorance (though it is interesting I chose basically the same implementation as mentioned in #103).

This PR is closed though; I reimplemented in #296 

Perhaps it will help if I add an example of the schema we are testing.

cec:
@pkarman I'm sure that adding a schema will help others see the arguments behind this proposal, which I agree with completely.

I will join the discussion on #296 , which now I understand to be the one receiving some attention.
I'll post an example schema of mine too.

Thank you!

PR #301, opened 2016-01-28
Addressable gem dependency fix

jsonenthusiast:
A new version of addressable gem is out (2.4.x). Without this fix, json-schema will not work with the latest version of the addressable gem.


iainbeeston:
I'm going to close this now #312 has been merged

PR #302, opened 2016-01-31
Convert readme to markdown

lencioni:
More people are more familiar with this format, and other documents in this repo are in markdown so it seems like a good reason to switch. Additionally, we can get syntax highlighting on github by using fenced code blocks in markdown, which I added in this PR. Rubydoc.info can also handle GitHub flavored markdown now, so everyone wins!


RST-J:
I don't mind. If everything stays the same (I think I saw broken doc from markdown somewhere once), I actually prefer this.

iainbeeston:
Me too

@lencioni Can you please squash your commits into one?

lencioni:
Sqauashed!

iainbeeston:
Great, thanks!

PR #313, opened 2016-02-29
Attach URI to Read Error

benfalk:
When getting a `404` or `500` error it was impossible to know which
file was causing the problem with a schema that had many external
references.  This attaches the uri that failed to the error if it's
an OpenURI error, which makes is easy to know what refrenced file
was the culprit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+# Changed
+- Attached failed `uri` or `pathname`  to read errors for more meaning
+
 ## [2.6.1] - 2016-02-26
 
 ### Fixed
--- a/lib/json-schema/schema/reader.rb
+++ b/lib/json-schema/schema/reader.rb
@@ -3,9 +3,8 @@ require 'pathname'
 
 module JSON
   class Schema
-    # Raised by {JSON::Schema::Reader} when one of its settings indicate
-    # a schema should not be readed.
-    class ReadRefused < StandardError
+    # Base for any reading exceptions encountered by {JSON::Schema::Reader}
+    class ReadError < StandardError
       # @return [String] the requested schema location which was refused
       attr_reader :location
 
@@ -15,7 +14,34 @@ module JSON
       def initialize(location, type)
         @location = location
         @type = type
-        super("Read of #{type == :uri ? 'URI' : type} at #{location} refused!")
+        super(error_message)
+      end
+
+      private
+
+      def error_message
+        "Read error for #{type_string} at #{location}!"
+      end
+
+      def type_string
+        type == :uri ? 'URI' : type.to_s
+      end
+    end
+
+    # Raised by {JSON::Schema::Reader} when one of its settings indicate
+    # a schema should not be readed.
+    class ReadRefused < ReadError
+      private
+      def error_message
+        "Read of #{type_string} at #{location} refused!"
+      end
+    end
+
+    # Raised by {JSON::Schema::Reader} when an attempt to read a schema fails
+    class ReadFailed < ReadError
+      private
+      def error_message
+        "Read of #{type_string} at #{location} failed!"
       end
     end
 
@@ -58,6 +84,7 @@ module JSON
       # @raise [JSON::Schema::ReadRefused] if +accept_uri+ or +accept_file+
       #   indicated the schema could not be read
       # @raise [JSON::Schema::ParseError] if the schema was not a valid JSON object
+      # @raise [JSON::Schema::ReadFailed] if read fails
       def read(location)
         uri  = JSON::Util::URI.parse(location.to_s)
         body = if uri.scheme.nil? || uri.scheme == 'file'
@@ -98,6 +125,8 @@ module JSON
         else
           raise JSON::Schema::ReadRefused.new(uri.to_s, :uri)
         end
+      rescue OpenURI::HTTPError, SocketError
+        raise JSON::Schema::ReadFailed.new(uri.to_s, :uri)
       end
 
       def read_file(pathname)
@@ -106,6 +135,8 @@ module JSON
         else
           raise JSON::Schema::ReadRefused.new(pathname.to_s, :file)
         end
+      rescue Errno::ENOENT
+        raise JSON::Schema::ReadFailed.new(pathname.to_s, :file)
       end
     end
   end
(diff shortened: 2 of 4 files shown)


lib/json-schema/schema/reader.rb
```diff
@@ -98,6 +98,8 @@ def read_uri(uri)
         else
           raise JSON::Schema::ReadRefused.new(uri.to_s, :uri)
         end
+      rescue OpenURI::HTTPError => error
```
iainbeeston: I'd prefer if we raised a custom error class here, eg. `JSON::Schema::ReadFailed` rather than re-raising the error of a third-party library (in case we ever changed what we use to load external files)

lib/json-schema/schema/reader.rb
```diff
@@ -98,6 +98,8 @@ def read_uri(uri)
...
+        raise error, "#{error} #{uri}"
       end
 
       def read_file(pathname)
```
iainbeeston: We should also rescue `Errno::ENOENT` for file read errors, so they behave in the same way.

VERSION.yml
```diff
@@ -1,3 +1,3 @@
 major: 2
 minor: 6
-patch: 1
```
iainbeeston: Please don't change VERSION.yml - that should only be done as part of a release

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Please keep to the changelog format described on [keepachangelog.com](http://kee
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+- Attached failed `uri` to OpenURI read errors for more meaning
```
iainbeeston: Can you please put this in a `### Changed` section?

iainbeeston:
Can you add some tests around this? (It should be easy to do, given an invalid uri)

benfalk:
Thanks for the feedback @iainbeeston !  I'll work on these :smile:

benfalk:
@iainbeeston I have made some edits, let me know what you think

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Made sure schemas of an unrecognized type raise a SchemaParseError (not Name error)
 
 ### Changed
+- Attached failed `uri` or `pathname`  to read errors for more meaning
```
iainbeeston: Sorry, that should be

    # [Unreleased]

    # Changed
    - Attached failed `uri` or `pathname`  to read errors for more meaning

    # [2.6.1] - 2016-02-26

(right now you've put your change under the 2.6.1 release, which has already been published

lib/json-schema/schema/reader.rb
```diff
@@ -3,19 +3,37 @@
...
+        @location = location
+        @type = type
+        super("Read of #{type_string} at #{location} refused!")
+      end
+    end
+
+    # Raised by {JSON::Schema::Reader} when an attempt to read a schema fails
+    class ReadFailed < ReadError
       def initialize(location, type)
         @location = location
```
iainbeeston: It feels like the parent initializer should set location and type, as type string depends on it, and the code is shared between ReadRefused and ReadFailed

test/test_bad_schema_ref.rb
```diff
@@ -19,7 +19,7 @@ def test_bad_uri_ref
     }
 
     data = [1,2,3]
-    assert_raises(Errno::ENOENT) do
+    assert_raises(JSON::Schema::ReadFailed) do
```
iainbeeston: Nice - I hadn't realised we were already catching these

iainbeeston:
Yeah that's looking good. Just another couple of changes. Would also be good to get @RST-J to check this too before we merge

benfalk:
@iainbeeston I've re-formatted the errors around a bit, let me know what you think.  Thanks!

lib/json-schema/schema/reader.rb
```diff
@@ -15,7 +14,34 @@ class ReadRefused < StandardError
...
         @type = type
-        super("Read of #{type == :uri ? 'URI' : type} at #{location} refused!")
+        super(error_message)
+      end
+
+      private
+
+      def error_message
```
RST-J: That is never read from the base class or did I overlook something? Wouldn't it be reasonable to require each subclass to implement a custom error message?

lib/json-schema/schema/reader.rb
```diff
@@ -58,6 +84,7 @@ def initialize(options = {})
       # @raise [JSON::Schema::ReadRefused] if +accept_uri+ or +accept_file+
       #   indicated the schema could not be read
       # @raise [JSON::Schema::ParseError] if the schema was not a valid JSON object
+      # @raise [JSON::Schema::ReadFailed] if read fails
```
RST-J: I'd extend the comment to also cover that this only occurs if reading the schema is permitted but the actual attempt to retrieve it fails.
As it is now, it appears to me like an explanation for wet water, telling it is wet :wink:

lib/json-schema/schema/reader.rb
```diff
@@ -15,7 +14,34 @@ class ReadRefused < StandardError
       def initialize(location, type)
```
RST-J: I think this could be DRYed up by refactoring the constructor to just take the object which could not be read - `@location` is always set to `location_object.to_s` and `@type` is `:uri` if the object is a URI, `:file` otherwise.
Just saw that, not definitely sure about it. What do you think?

test/test_bad_schema_ref.rb
RST-J: I think it would be good to test that `location` and `type` are set correctly in both error cases. @iainbeeston do you have an idea how this could be easily integrated in our test helper setup?

RST-J:
I added my feedback - just minor things, in general I like this improvement. I can that errors like this are currently not much fun during development.

iainbeeston:
Hey @benfalk do you have time to implement the few points of feedback?

benfalk:
@iainbeeston yup, sorry for dropping off on this.  Been using my fork and this fell off on my radar.  I'll get this updated a little later today.

iainbeeston:
Hey @benfalk - just want to poke you over this, I'd love to merge your PR

iainbeeston:
Or let me know if you're happy for me to fix it and merge it myself (I'm happy either way)

iainbeeston:
I've fixed up the last few things and merged manually in 3f07ef0599c9a0e080840bfd1be04df433b7bbff

iainbeeston:
Thanks @benfalk

PR #322, opened 2016-04-15
Made it possible to have refs that include URI-special characters

### LINKED ISSUE #319: JSON pointers are broken when they contain a `:`
If I read [RFC 6901](https://tools.ietf.org/html/rfc6901) correctly (and I'm slightly RFC impaired), the colon character `:` should be OK to use in JSON pointers. However, `json-schema-2.6.1` will break, check out this [ready to run gist](https://gist.github.com/jayniz/e8849ea528af6d205698).

The RFC states:

```
 The ABNF syntax of a JSON Pointer is:

      json-pointer    = *( "/" reference-token )
      reference-token = *( unescaped / escaped )
      unescaped       = %x00-2E / %x30-7D / %x7F-10FFFF
         ; %x2F ('/') and %x7E ('~') are excluded from 'unescaped'
      escaped         = "~" ( "0" / "1" )
        ; representing '~' and '/', respectively
```

So a `:` in a JSON pointer should be OK whereas `/` and `~` must be escaped:

```ruby
> !!":"[/[\u0000-\u002E\u0030-\u007d\u007F-\u10FFFF]/] # : is OK
=> true
> !!"~"[/[\u0000-\u002E\u0030-\u007d\u007F-\u10FFFF]/] # ~ must be escaped
=> false
> !!"/"[/[\u0000-\u002E\u0030-\u007d\u007F-\u10FFFF]/] # / must be escaped
=> false
>
```

If you run the [gist](https://gist.github.com/jayniz/e8849ea528af6d205698), though, you'll see something like 

```
~/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/json-schema-2.6.1/lib/json-schema/attributes/ref.rb:63:in `block in get_referenced_uri_and_schema': The fragment '/definitions/thing' does not exist on schema af6a3d4a-9191-5753-9d90-edb43a637c46 (JSON::Schema::SchemaError)
```

If you paste the schema and data examples from the gist into an online JSON Schema tool like http://jsonschemalint.com/draft4/ it works as expected.

iainbeeston:
Turns out that before:

* if you had a ref that included a colon, the text before the colon was
  treated as the scheme and the text after the colon as a path
* the unescape_uri method was actually only parsing the path of the
  uri (therefore anything before a colon was being stripped when passed
  to unescape_uri)

Fixes #319

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+- Made it possible to include colons in a $ref
+
 ### Changed
 - Reformatted examples in the readme
 
--- a/lib/json-schema/attributes/ref.rb
+++ b/lib/json-schema/attributes/ref.rb
@@ -48,11 +48,11 @@ module JSON
         if ref_schema
           # Perform fragment resolution to retrieve the appropriate level for the schema
           target_schema = ref_schema.schema
-          fragments = temp_uri.fragment.split("/")
+          fragments = JSON::Util::URI.parse(JSON::Util::URI.unescape_uri(temp_uri)).fragment.split("/")
           fragment_path = ''
           fragments.each do |fragment|
             if fragment && fragment != ''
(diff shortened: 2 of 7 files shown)


jayniz:
🎉

RST-J:
:+1:

PR #323, opened 2016-04-15
fix fragment when used with extended schema

dpatti:
This refactors some of `Validator` to fix an issue with validating a
fragment of an extended schema against a json object. The problem is
that while navigating through the schema, we are using the configured
`:version` option as the validator for each partial schema. This is an
issue because if the root of the schema declares a `$schema`, that
information is lost in the traversal.

The refactor does two things. Firstly, it does not write back into
`@options[:version]` and it stops using it outside of the constructor
entirely. The only place that should know about this is the initialized
schema, and from that point on, the canonical validator should be the
schema object's validator since it makes a decision about which to use
in its constructor. So we can reduce complexity by using that validator
in the `:validate_schema` option branch.

The other place that used `:version` is the `schema_from_fragment`
method, which is where the bug exists. What we're doing here, instead of
dealing with schemas _and_ hashes during the iterator, is pulling the
schema object out, digging through it using the fragments, and then
wrapping the result in a schema using the `base_schema`'s validator.

Tests have been included that demonstrate the problem.

### DIFF
--- a/lib/json-schema/validator.rb
+++ b/lib/json-schema/validator.rb
@@ -41,8 +41,7 @@ module JSON
       @options = @@default_opts.clone.merge(opts)
       @errors = []
 
-      validator = JSON::Validator.validator_for_name(@options[:version])
-      @options[:version] = validator
+      configured_validator = JSON::Validator.validator_for_name(@options[:version])
       @options[:schema_reader] ||= JSON::Validator.schema_reader
 
       @validation_options = @options[:record_errors] ? {:record_errors => true} : {}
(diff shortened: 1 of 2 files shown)


iainbeeston:
So what you're saying is that if we validate against a fragment, we should use the version from the base schema. That makes sense. However, I think your test is invalid - I think `odd-a` should have `type: "object"`?

dpatti:
I don't know if `type: "object"` is specifically necessary, and the other tests in the file do not have it. I am happy to add it if you think it is important. As it stands now, the test should fall without the patch and passes with it.

iainbeeston:
I'll take a deeper look at this shortly. I think the essence of what you're
doing is right.

PR #332, opened 2016-06-24
Stopped the additional properties test from using class_eval

iainbeeston:
`class_eval` is dangerous to use, as it injects a string into a class as
ruby code, meaning that it's very easy to make mistakes. In this case
there is no need to use it, and I have re-written this test to use
multiple assertions per test instead (also not ideal, but better).

I've also made this test use the same assertion helpers as every other
test (previously it was defining it's own assertion helpers).

### DIFF
--- a/test/extends_nested_test.rb
+++ b/test/extends_nested_test.rb
@@ -1,50 +1,41 @@
 require File.expand_path('../support/test_helper', __FILE__)
 
 class ExtendsNestedTest < Minitest::Test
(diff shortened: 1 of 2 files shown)


iainbeeston:
Hey @RST-J am I ok to merge this? It seems innocuous but it will make refactoring other tests easier

RST-J:
Yeah, go! :smile: :+1:

iainbeeston:
Done, thanks!

PR #333, opened 2016-06-24
Use self or self.class rather than fully qualified class names

iainbeeston:
In validator.rb and schema.rb there are many places where we refer to
the current class explicitly, by it's fully qualified
name (eg. `JSON::Schema`) where instead we could be using `self` or
`self.class`. I believe using `self` is cleaner and more obvious, so I
propose we change the code to use self.


RST-J:
It actually changes (I'd say improves) one thing, which I guess isn't really important however: If you have a subclass for which these methods are overridden, it uses the overridden implementation instead of the fully qualified base class' one.
I agree with the  better readability and hence :+1:

PR #334, opened 2016-06-24
Fixed issues with caching

iainbeeston:
I discovered that the flag we have to clear the cache after validation (`:clear_cache => true`) didn't actually work. I've added tests around that and fixed the problem. There were also issues with `JSON::Validator.cache_schemas`.

There was also some minor refactoring of the test helpers, to make clearing the cache and validating schemas the default, which wouldn't have worked before merging this.

### DIFF
--- a/lib/json-schema/validator.rb
+++ b/lib/json-schema/validator.rb
@@ -18,7 +18,7 @@ module JSON
   class Validator
 
     @@schemas = {}
-    @@cache_schemas = false
+    @@cache_schemas = true
     @@default_opts = {
       :list => false,
       :version => nil,
@@ -48,7 +48,7 @@ module JSON
       @validation_options = @options[:record_errors] ? {:record_errors => true} : {}
       @validation_options[:insert_defaults] = true if @options[:insert_defaults]
       @validation_options[:strict] = true if @options[:strict] == true
-      @validation_options[:clear_cache] = true if @@cache_schemas || @options[:clear_cache]
+      @validation_options[:clear_cache] = true if !@@cache_schemas || @options[:clear_cache]
 
       @@mutex.synchronize { @base_schema = initialize_schema(schema_data) }
       @original_data = data
--- a/lib/json-schema/validator.rb
+++ b/lib/json-schema/validator.rb
@@ -26,7 +26,7 @@ module JSON
       :record_errors => false,
       :errors_as_objects => false,
       :insert_defaults => false,
-      :clear_cache => true,
+      :clear_cache => false,
       :strict => false,
       :parse_data => true
     }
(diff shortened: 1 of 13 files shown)


test/support/number_validation.rb
```diff
@@ -78,16 +78,5 @@ def test_multiple_of
       # other types are disregarded
       assert_valid schema, {'a' => 'hi'}
     end
-
-    def test_multiple_of_zero
```
iainbeeston: The schema for this test was actually invalid (`multipleOf` must be greater than 0)

test/caching_test.rb
```diff
@@ -0,0 +1,61 @@
...
+    schema = JSON::Schema.new(initial_schema, URI)
+    JSON::Validator.add_schema(schema)
+    assert_valid(URI, "foo")
+
+    initial_schema = {
+      "type" => "number"
+    }
+    schema = JSON::Schema.new(initial_schema, URI)
+    JSON::Validator.add_schema(schema)
+    assert_valid(URI, 123)
```
iainbeeston: The logic here is wrong - when cache_schemas is true this test should fail because the previous one should still be cached

test/caching_test.rb
```diff
@@ -0,0 +1,61 @@
...
+    schema = JSON::Schema.new(initial_schema, URI)
+    JSON::Validator.add_schema(schema)
+    assert_valid(URI, "foo", :clear_cache => false)
+
+    initial_schema = {
+      "type" => "number"
+    }
+    schema = JSON::Schema.new(initial_schema, URI)
+    JSON::Validator.add_schema(schema)
+    assert_valid(URI, "foo")
```
iainbeeston: This would be clearer if it used refute_valid and used 123 as data (not a string)

iainbeeston:
@@cache_schemas should probably default to true, as that's the current behaviour

Also, there should probably be some info in the readme about caching

lib/json-schema/validator.rb
```diff
@@ -48,7 +48,7 @@ def initialize(schema_data, data, opts={})
       @validation_options = @options[:record_errors] ? {:record_errors => true} : {}
       @validation_options[:insert_defaults] = true if @options[:insert_defaults]
       @validation_options[:strict] = true if @options[:strict] == true
-      @validation_options[:clear_cache] = false if @options[:clear_cache] == false
+      @validation_options[:clear_cache] = true if @@cache_schemas || @options[:clear_cache]
```
iainbeeston: The logic here should be:

    !@@cache_schemas || @options[:clear_cache]

lib/json-schema/validator.rb
```diff
@@ -18,15 +18,15 @@ module JSON
   class Validator
 
     @@schemas = {}
-    @@cache_schemas = false
+    @@cache_schemas = true
```
iainbeeston: The default behaviour previously was to cache schemas, even though the `@@cache_schemas` variable was false

RST-J:
:+1:

PR #315, opened 2016-03-11
Updated example of extending an existing schema

sgravrock:
The README gives an example of how to validate an extended schema, but it looks like the code is out of date. This fixes it.

PR #335, opened 2016-06-28
Deprecated #extend_schema_definition

iainbeeston:
There are two ways right now to extend a schema, one is by subclassing
either an existing validator (eg. `JSON::Schema::Draft4`) and the other is
by subclassing the base validator class (`JSON::Schema::Validator`) and
calling `extend_schema_definition` in the initializer. This method
merges the recognised json schema properties of another validator into
your new validator. Unfortunately this is effectively the same as
subclassing. What's worse, `extend_schema_definition` does not merge in
any of the other characteristics of a validator (eg. the formats or
names etc needed for a schema to work). All of the validator classes
that come with json-schema use subclassing, rather than
`extend_schema_definition`.

Because of this, I propose that we should deprecate
`extend_schema_definition` and encourage people to subclass instead.


iainbeeston:
Incidentally this also includes the fix from #315

PR #337, opened 2016-06-29
Added a test to ensure that code in the readme is valid

iainbeeston:
With this PR, if any code in the readme raises an error, the test suite
will fail.


iainbeeston:
I'm closing this, it was a nice idea but for now it's too much trouble for what it's worth

PR #338, opened 2016-06-29
Made sure all validate methods go through the same call chain

iainbeeston:
Right now when a schema is validated, we perform a different call path
depending on which of the validate methods is called. Schema validation
has it's own path too.

I've changed this so everything is validated via
`JSON::Validator#validate!`, all other validate methods ultimately call
that. This means that the varied validate class methods on `Validator`
are simplified and just add or remove validation options.


RST-J:
:+1: +2 actually, I like tidy code.

PR #340, opened 2016-06-29
Removed the ruby 1.8 build from travis

iainbeeston:
Ruby 1.8 support hasn't worked for some time, and we've been ignoring
build failures on travis for ruby 1.8. Here are some examples of code
that has already been released, which is incompatible with ruby 1.8:

* DateTime.rfc3339 was only introduced in ruby 1.9
* Hash#select was changed to return a Hash in ruby 1.9
* Hash keys became ordered in ruby 1.9

I haven't added anything to the changelog because this isn't a new
thing, it's been the reality for some time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: "ruby"
 
 rvm:
-  - "1.8"
   - "1.9"
   - "2.0"
   - "2.1"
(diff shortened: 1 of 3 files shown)


RST-J:
:+1:

PR #341, opened 2016-06-29
Made tests use helpers where possible

iainbeeston:
Changed tests to use the validation helpers where possible, to simplify
the test suite and make more of the tests behave in the same way.


iainbeeston:
This is a little harder than I had expected because of dependencies in the test suite... This PR isn't quite ready yet

iainbeeston:
Closing this as it needs work

PR #345, opened 2016-06-30
Moved Validator#absolutized_uri and RefAttribute's ref parsing into the URI module

iainbeeston:
I just noticed that we have this URI processing method in Validator, and
similar code in RefAttribute. I had thought we'd moved all of these into
the URI module, but these must have been missed.

I've taken the opportunity to tidy them up slightly too.

One day we might be able to unify these methods

### DIFF
--- a/lib/json-schema/util/uri.rb
+++ b/lib/json-schema/util/uri.rb
@@ -23,6 +23,43 @@ module JSON
         normalized_uri
       end
 
+      def self.absolutize_ref(ref, base)
+        ref_uri = strip_fragment(ref.dup)
+
+        return ref_uri if ref_uri.absolute?
+        return parse(base) if ref_uri.path.empty?
+
+        uri = strip_fragment(base.dup).join(ref_uri.path)
+        normalized_uri(uri)
+      end
+
+      def self.normalize_ref(ref, base)
+        ref_uri = parse(ref)
+        base_uri = parse(base)
+
+        ref_uri.defer_validation do
+          if ref_uri.relative?
+            ref_uri.merge!(base_uri)
+
+            # Check for absolute path
+            path, fragment = ref.to_s.split("#")
+            if path.nil? || path == ''
+              ref_uri.path = base_uri.path
+            elsif path[0,1] == "/"
+              ref_uri.path = Pathname.new(path).cleanpath.to_s
+            else
+              ref_uri.join!(path)
+            end
+
+            ref_uri.fragment = fragment
+          end
+
+          ref_uri.fragment = "" if ref_uri.fragment.nil? || ref_uri.fragment.empty?
+        end
+
+        ref_uri
+      end
+
       def self.parse(uri)
         if uri.is_a?(Addressable::URI)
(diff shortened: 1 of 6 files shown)


iainbeeston:
@RST-J This PR also helps to move all URI parsing code into one place

test/uri_util_test.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    assert_equal Addressable::URI.parse('http://www.example.com/foo/hello/world#'), JSON::Util::URI.normalize_ref(uri, base)
+    assert_equal Addressable::URI.parse('http://www.example.com/foo/hello/world#'), JSON::Util::URI.absolutize_ref(uri, base)
+  end
+
+  def test_ref_addressable_uri_with_host
+    uri = Addressable::URI.new(:host => 'foo-bar.com')
+    base = 'http://www.example.com/foo/#bar'
+
+    assert_equal Addressable::URI.parse('http://www.example.com/foo-bar.com#'), JSON::Util::URI.normalize_ref(uri, base)
```
RST-J: Is that really the expected behaviour, appending the host as path? And if so, is that because of the URI lacking a protocol?

lib/json-schema/util/uri.rb
```diff
@@ -23,6 +23,46 @@ def self.normalized_uri(uri, base_path = Dir.pwd)
         normalized_uri
       end
 
+      def self.absolutize_ref(ref, base)
+        ref_uri = strip_fragment(ref.dup)
+
+        if ref_uri.absolute?
```
RST-J: The [style guide](https://github.com/bbatsov/ruby-style-guide#no-nested-conditionals) suggests using guard clauses as in the refactored code. I personally prefer them (in our team we decided to usually follow the style guide).
But that's no religious comment, I just want to put that on the table.

test/uri_util_test.rb
```diff
@@ -0,0 +1,75 @@
...
+    uri = 'http://foo-bar.com/some/thing'
+    base = 'http://www.example.com/foo/#bar'
+
+    assert_equal Addressable::URI.parse('http://foo-bar.com/some/thing#'), JSON::Util::URI.normalize_ref(uri, base)
+    assert_equal Addressable::URI.parse('http://foo-bar.com/some/thing#'), JSON::Util::URI.absolutize_ref(uri, base)
+  end
+
+  def test_ref_uri_with_fragment
+    uri = 'http://foo-bar.com/some/thing#foo'
+    base = 'http://www.example.com/foo/#bar'
```
RST-J: I think this test case and the one below would have improved readability if the fragment of the first and the path of the second URI would not both be `foo`.

test/uri_util_test.rb
iainbeeston: I've just tried to document the existing behaviour for now. I've added some extra tests that do include a schema and this changes the behaviour entirely!

iainbeeston:
@RST-J I've corrected the issues you've mentioned, thanks for the feedback!

PR #347, opened 2016-07-13
Update 3.x to include current master

RST-J:
I merged master into our very stale 3.x branch. There are some spots I'm not sure about, I'll comment them directly (it was only one).
@iainbeeston In general we should come up with a decision about how to go on with v3. What is your current opinion on that?


lib/json-schema/validator.rb
```diff
@@ -445,13 +448,13 @@ def initialize_schema(schema)
...
           # Build a fake URI for this
-          schema_uri = Addressable::URI.parse(fake_uuid(schema))
+          schema_uri = JSON::Util::URI.parse(fake_uuid(schema))
           schema = JSON::Schema.new(JSON::Validator.parse(schema), schema_uri, @options[:version])
           if @options[:list] && @options[:fragment].nil?
             schema = schema.to_array_schema
           end
           Validator.add_schema(schema)
-        rescue
+        rescue JSON::ParserError
```
RST-J: Currently in master we have [`JSON::Schema::JsonParseError`](https://github.com/ruby-json-schema/json-schema/blob/master/lib/json-schema/validator.rb#L541) being rescued here. But if we only support JSON, I think there is no need to wrap parsing errors any more. Did I miss something?

RST-J:
The [1.9 job](https://travis-ci.org/ruby-json-schema/json-schema/jobs/144604611) failed because it tries to install a too new json gem version. Does someone now how fix that?

iainbeeston:
I think that the 3.0 branch is defunct - the two things that were in 3.0 that weren't in master were the removal of ruby 1.8 support, and removing support for alternative json back-ends. We've already removed ruby 1.8 support (in #340 ) and I've extracted out @pd 's commit to remove json backends in #339 . I think that once #339 is merged and released, we can close this PR and delete the 3.0 branch.

I think more generally we should continue making refactoring in master in minor releases, and once we're ready to make breaking changes (like removing caching and class variables) we should do that in a new 3.0 release. But we're not ready for that yet and maintaining a 3.0 branch is hard work.

iainbeeston:
@RST-J I just noticed the ruby 1.9 build issue too - I'll fix that

RST-J:
I thought, we already had more in 3.0, but under this circumstances I agree with you.
So we need two phases, the first for cleaning up the code while remaining backwards compatible and then at some point we need to introduce all the changes we think are important for 3.0. That actually could help focusing on the respective goal, rather than doing both at the same time or interleaved.
Nevertheless, when we start developing/merging the breaking changes, this should ideally be within a relativly short time frame or we'll run in the same or a similar problem again.

iainbeeston:
I agree. Can we close this pull request for now?

I should probably delete the 3.0 branch too - until (as you said) we're ready to make a short set of breaking changes for 3.0.

PR #348, opened 2016-07-25
Item partial tuples

jlfaber:
Currently, when 'items' is an array of schemas, validation will fail if the target array has fewer elements than the 'items' array.  This conflicts with the spec, which allows for partial (or even empty) tuples to pass validation unless 'minItems' is also set.

See the relevant section of [Understanding JSON Schema](https://spacetelescope.github.io/understanding-json-schema/reference/array.html#tuple-validation) for examples of this.  Note specifically the statement: "It's okay to not provide all the items."

This PR modifies the array_validation tests to check for this condition and fixes the corresponding code in the attributes directory.

### DIFF
--- a/lib/json-schema/attributes/items.rb
+++ b/lib/json-schema/attributes/items.rb
@@ -16,7 +16,7 @@ module JSON
 
         when Array
           items.each_with_index do |item_schema, i|
-            next unless i < data.length
+            break if i >= data.length
             schema = JSON::Schema.new(item_schema, current_schema.uri, validator)
             schema.validate(data[i], fragments + [i.to_s], processor, options)
           end
--- a/lib/json-schema/attributes/items.rb
+++ b/lib/json-schema/attributes/items.rb
@@ -16,6 +16,7 @@ module JSON
 
         when Array
           items.each_with_index do |item_schema, i|
+            next unless i < data.length
             schema = JSON::Schema.new(item_schema, current_schema.uri, validator)
             schema.validate(data[i], fragments + [i.to_s], processor, options)
           end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Made sure we really do clear the cache when instructed to
 - It's now possible to use reserved words in property names
 - Removed support for setting "extends" to a string (it's invalid json-schema - use a "$ref" instead)
+- Relaxed 'items' and 'allowedItems' validation to permit arrays to pass even
+  when they contain fewer elements than the 'items' array.  To require full tuples,
+  use 'minItems'.
 
 ### Changed
 - Made all `validate*` methods on `JSON::Validator` ultimately call `validate!`
--- a/lib/json-schema/attributes/additionalitems.rb
+++ b/lib/json-schema/attributes/additionalitems.rb
@@ -11,7 +11,7 @@ module JSON
 
         case schema['additionalItems']
         when false
-          if schema['items'].length != data.length
+          if schema['items'].length < data.length
             message = "The property '#{build_fragment(fragments)}' contains additional array elements outside of the schema when none are allowed"
             validation_error(processor, message, fragments, current_schema, self, options[:record_errors])
           end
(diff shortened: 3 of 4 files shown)


jlfaber:
CI failing due to an issue fixed by #349.

jlfaber:
CI now passes.  Also updated the common test suite ref.
@iainbeeston, this should be ready to go.

lib/json-schema/attributes/items.rb
```diff
@@ -16,6 +16,7 @@ def self.validate(current_schema, data, fragments, processor, validator, options
 
         when Array
           items.each_with_index do |item_schema, i|
+            next unless i < data.length
```
RST-J: Couldn't this be replaced by `break if i >= data.length``?

RST-J:
:+1:

rymndhng:
@jlfaber any chance you can fix up this PR -- I'd love to get this fix in :)

jlfaber:
I've rebased, but this is now failing (again) due to another unrelated issue.

iainbeeston:
It looks like something changed in the [common test suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite), that's also causing master to fail. I'll take a closer look at that issue

iainbeeston:
@jlfaber I've figured out why the tests are failing on master. Once #360 is merged, and you rebase on top of that, the tests here should pass

RST-J:
#360 is merged

jlfaber:
@iainbeeston tests now passing -- ready to merge.  Thx for the fix in #360 !

iainbeeston:
Great - I'll take one last look at this today and hopefully get it merged and released

iainbeeston:
Thanks @jlfaber I've merged that, and I'll try to get it released soon.

Interestingly the original json-schema spec is very vague about how this should work. I'm going to add some extra tests to the common test suite so we can be sure we are doing the right thing

iainbeeston:
My PR on the common test suite is https://github.com/json-schema-org/JSON-Schema-Test-Suite/pull/127

jlfaber:
Thanks @iainbeeston for the merge.

The current schema validation spec (http://json-schema.org/latest/json-schema-validation.html#anchor37) is pretty specific about how items should work.  Note the example 5.3.1.3 which shows this schema:
```json
{
  "items": [ {}, {}, {} ],
  "additionalItems": false
}
```
and these examples of valid instances:
```json
[]
[ [ 1, 2, 3, 4 ], [ 5, 6, 7, 8 ] ]
[ 1, 2, 3 ]
```
The valid arrays have zero, two, and three elements, respectively.

iainbeeston:
Yes that's true, but what I was thinking of was that it doesn't explain how the schemas in `items` should be used to validate the contents of the array. Here's the best I can find:

> if the value of "additionalItems" is boolean value false and the value of "items" is an array, the instance is valid if its size is less than, or equal to, the size of "items". 

It seems safe to assume that each consecutive array element should match the schema in items at the same position, but then what happens when there are more elements in the array than there are items in the schema? (only relevant if `additionalItems` is `false`)

jlfaber:
Did you mean when ```additionalItems``` is ```true```?  If it's false (meaning additional items are *not* allowed) and there are more elements in the array than there are in the schema, then that's a clear failure.

This section (http://json-schema.org/latest/json-schema-validation.html#anchor131) offers a bit more guidance about validating the elements of an array against an ```items``` schema:

> 8.2.3.2.  If "items" is an array
>   In this situation, the schema depends on the index: if the index is less than, or equal to, the size of "items", the child instance must be valid against the corresponding schema in the "items" array; otherwise, it must be valid against the schema defined by "additionalItems".

So to summarize: ```items``` can be either a single schema object, which is used to validate all elements of the subject array, or an array of schema objects, each of which is used to validate the corresponding element of the subject array _if it exists_.  The ```items``` attribute itself imposes no restrictions on the length of the subject array.

When ```items``` is not present or is a single schema object, ```additionalItems``` is disregarded.  

When ```items``` is an array, ```additionalItems``` may be either ```true``` (the default, which imposes no restrictions at all), ```false``` (which requires that the subject array length be no larger than the length of the ```items``` array), or a single schema object (which is used to validate any subject array elements that do not correspond to a schema in the ```items``` array).

PR #349, opened 2016-07-26
Remove ruby 1.9 from travis CI spec

jlfaber:
As of July 1, the latest version of the json gem (2.0.1) now requires ruby ~> 2.0.  This is breaking CI testing on master.  This PR removes the ruby 1.9 entry from the travis spec file.


iainbeeston:
Sorry, but we do want to keep ruby 1.9 support. The correct solution to this (which I haven't had time to do yet) is to use a custom gemfile on the ruby 1.9 build which limits the json gem to versions less than 2.0 (which raised the minimum ruby version to 2.0)

iainbeeston:
#350 should resolve this issue.

jlfaber:
Ah. Well done, then.  Thanks.

PR #350, opened 2016-07-27
Restricted the ruby 1.9 build to json < 2

iainbeeston:
Version 2.0 of the json gem requires ruby 2+. We should continue
supporting ruby 1.9, but to do so we need to make sure that travis uses
version 1 of the json gem.

I've done this by adding a custom gemfile for ruby 1.9 that restricts
json to version 1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: "ruby"
 
 rvm:
-  - "1.9"
   - "2.0"
   - "2.1"
   - "2.2"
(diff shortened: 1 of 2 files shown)


iainbeeston:
@RST-J Could I please get a quick code review of this? (Right now the build is broken on master because of a change in the latest version of the json gem)

RST-J:
You can :smiley: :+1:

iainbeeston:
@RST-J Great, thanks!

PR #359, opened 2016-09-28
Use latest json-schema draft in tests by default

iainbeeston:
Lots of core tests were explicitly using draft-03, which works slightly
differently to the latest (draft-04 as I write this). I've rewritten
them to not specify the schema version at all, so they default to using
the latest.

To me this feels safer, as (unless explicitly specified) we're always
testing against the default draft.

The main change here is the way that the "required" property has
changed (moving from being defined on the definition of each "property"
to being defined alongside the "properties" object).


RST-J:
I agree.

PR #360, opened 2016-09-28
Made it possible to have a property named "$ref"

### LINKED ISSUE #357: Test failure due to update in JSON-Schema-Test-Suite
This PR (which has been merged) introduced the failing tests.

https://github.com/json-schema-org/JSON-Schema-Test-Suite/pull/125

iainbeeston:
For a few weeks we've had a test from the json-schema common test suite
failing on master. The test was checking that it was possible to
validate a property named "$ref" (and that it wasn't reserved or
interpreted as a json-reference).

The problem was that whenever we had a property name that has special
meaning in json-schema (ie. one that matches the keys of the
Schema::Validator#attributes hash), we'd skip a level in the schema
object while validating, which would cause errors. To fix the problem
I've made sure we always descend one level at a time (see
attributes/additionalproperties.rb line 17).

Fixing this caused new tests to fail, in test/extends_nested_test.rb. It
turns out that these tests were depending on the old incorrect
behaviour.

What's more, the same tests were using an incorrect form of the extends
property. According to the json-schema spec, extends should take a
json-schema, or an array of json-schemas. However, these tests were
passing a string to extends. It seems that although this is incorrect,
we had code in attributes/extends.rb which would automatically correct
this incorrect syntax (ie. from {"extends": "foo.json"} to {"extends:
{"$ref": "foo.json"}}). I've removed that workaround, as well as
correcting the tests.

Fixes #357


iainbeeston:
@RST-J could you please take a look at this when you have time? It fixes test suite failures on master, so we can start merging pull requests again

iainbeeston:
Awesome! Thanks for the fast review!

Issue #329, opened 2016-06-03
JSON::Util::URI.parse memory leak


skippy:
hey folks,

I benchmarked my app, which does JSON Schema validation in a tight loop (10ks of validations per minute per thread, 10m of validations per run), and I noticed the following:
* the single largest class of objects, by far, and across my entire code base, is Addressable::URI
* the Addressable::URI objects are not being GC'ed 

Using the (damn wonderful) `stackprof` gem, almost all the Addressable URI objects are created by this helper:  https://github.com/ruby-json-schema/json-schema/blob/master/lib/json-schema/util/uri.rb#L26.  500m+ Addressable URI objects are created per run, as my json-schemas use `$ref`s quite a bit.  

But what is odd is none of these objects are being GC'ed.  At a cursory glance of the code, it seems that they should be as `#dup` is used.  However, if I remove the class-level cache and the use of `#dup`, like so:

```ruby
module JSON
  module Util
    module URI

      def self.parse(uri)
        # do NOT parse a Addressable::URI; behind the scenes it creates a dup
        # which leaves a reference to the parent object (as it is not duping the 
        # string objects), which causes a memory explosion
        # 
        return uri if uri.is_a?(Addressable::URI)
        Addressable::URI.parse(uri)
      rescue Addressable::URI::InvalidURIError => e
        raise JSON::Schema::UriError.new(e.message)
      end

    end
  end
end
```

the memory leak is gone.  My process went from growing to ~ 1.5GB per run to staying at a consistent ~ 250MB.

I suspect the issue is the `@parse_cache` combined with the dup.  The `#dup` creates a new object, and within `Addressable::URI` it calls `#dup` on the strings.  However my understanding is that this is a copy-on-write process (see http://www.justskins.com/forums/memory-behavior-of-string-8195.html), which means that a reference will be kept around to the shared parent object until it is modified.  All it would take is one string reference from the parent object (which is never deleted as it is stored within `@parse_cache`) to stick around to prevent *any* of the dupped object from being GC'ed.  I haven't dived into this a ton or into the C code (outside of benchmarking and patching) but this is what I suspect is going on.  

[...]

PR #361, opened 2016-10-01
Made sure clear_cache also clears the cache of parsed uris

iainbeeston:
There are two main caches in json-schema. One is for schemas, one is for
parsed uris. Previously, setting `:clear_cache` as a validation option
would clear the schema cache but not the uri cache. This has been
reported to cause a memory leak (see issue #329) as over time the uri
cache gets bigger and bigger and is never cleared.

This change makes sure that whenever the schema cache is cleared, the
uri cache is also cleared.

Incidentally this change is not thread safe, and multithreaded
applications that clear the cache may see  issues. But until we can
refactor to either disable caching or get rid of class variables,
I don't believe there's much we can (realistically) do about that. This
is still an improvement on what's gone before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+- Made the `:clear_cache` option for `validate` also clear the URI parse cache
+
 ## [2.7.0] - 2016-09-29
 
 ### Fixed
--- a/lib/json-schema/util/uri.rb
+++ b/lib/json-schema/util/uri.rb
@@ -63,6 +63,11 @@ module JSON
 
         Addressable::URI.unescape(parsed_uri.path)
       end
+
+      def self.clear_cache
+        @parse_cache = {}
+        @normalize_cache = {}
+      end
     end
   end
 end
--- a/lib/json-schema/validator.rb
+++ b/lib/json-schema/validator.rb
@@ -299,6 +299,7 @@ module JSON
 
       def clear_cache
         @@schemas = {}
+        JSON::Util::URI.clear_cache
       end
 
       def schemas
--- a/test/uri_parsing_test.rb
+++ b/test/uri_parsing_test.rb
@@ -2,6 +2,16 @@
 require File.expand_path('../support/test_helper', __FILE__)
 
 class UriParsingTest < Minitest::Test
+  def populate_cache_with(str, &blk)
+    cached_uri = Addressable::URI.parse(str)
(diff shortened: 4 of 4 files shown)


RST-J:
My brain sometimes tickles me with the idea of a major restructuring of json-schema. Basically to extract functionality according to the separation of concerns principle. The validator base class does really, really much and I still have not the feeling to be at home there (which I consider a bad sign).
I have no clear idea whether and how to do that. I also haven't spent too much thoughts about it as this only makes sense, if we a) agree this is a good idea and b) then agree on how this then considered good idea should be carried out.

iainbeeston:
Yes I agree. I can see a way forward though by extracting out functionality one object at a time. There is clearly an object for URI parsing, and one for loading data from a file or uri. We should not be using class variables, but instead keep a set of objects internal to each validator object, and allow the user to reuse validator objects if they want to cache previously used data.

Another option that I have considered is extracting out some of the functionality into separate gems. For example, a gem could be made that just evaluates json references. Json parsing is already handled by the json gem. The json schema spec was split into three parts for draft 4 and these could possibly be represented by 3 gems? (Though I'm not sure if that would work?)

Regarding this PR, I think we should possibly have a single object that handles URI parsing, and (for now) keep an instance of that object in a class variable in Validator. That object could keep the URI caches internal to itself. This would be similar to the Reader objects we already have. Although I would argue that what is already in this pull request is good enough for now and we can improve the object design in a new pull request.

iainbeeston:
Oh, and another thing - I've been trying hard (with this and other pull requests) to improve the reliability and coverage of the test suite, so that if we do choose to refactor we can be more confident that nothing will break. Although we have a lot of tests some areas do not even have a single test that covers them! In other cases we have tests that only pass if they are run in the correct order!

RST-J:
Fine. I agree that we should take small steps - one aspect at a time. Actually I think this is the only realistic way to go. And I'm also in favour of improving the test suite itself and extending coverage along the way - wrt testing I'm rather the 'more is better' kind of guy, even more as this is a library.

What three parts are you referring to? I currently do not really have the spec (details) present in my mind - I'm probably kind of asked to much of by the 2nd paragraph in your first comment. ;)
In general I'd say we shouldn't make things too complex. Managing two dependencies to achieve one thing - validating json-schema is what users want to do - could introduce additional complexity. But if they make sense on their own, it could also be a good decision, although my gut feeling right now suggests otherwise.

And finally, yes, lets merge this. Its a small step.

iainbeeston:
Ah the three json schema drafts are the core, validation and hyper schema drafts. I suspect it was split up in draft4 because it was becoming too long...

Issue #254, opened 2015-05-14
Exceptions are used for control flow in non-bang methods


mjc:
```JSON::Validator.validate``` still uses exceptions internally for control flow.

This is a bug as exceptions have performance implications on MRI and dramatic performance implications on JRuby.  See [this PR](https://github.com/jruby/activerecord-jdbc-adapter/pull/557) for an example of a 9x speedup on JRuby from removing raise/rescue from the hot path in an already highly-optimized library.

I would be happy to start working on fixing this but it seems like a big undertaking and I want to be sure others agree this is worthwhile.

kwando:
@mjc it is a good idea change this like you suggest

iainbeeston:
Yes I'm totally behind this if you have a plan.

Also it might be worth a quick look at #239 where I've tried to clean up
the existing error handling

PR #362, opened 2016-10-02
Load local copy of draft schemas

iainbeeston:
If you refer to a draft schema in a $ref (as the common test suite does)
json-schema will load it from json-schema.org even though the gem
contains a local copy.

This change patches JSON::Validator#load_ref_schema to check whether a
$ref refers to a draft schema before trying to load from the web, and if
so parsing and serving up the local copy.

I've also replaced the local copy of draft3 with the exact version found
online (only whitespace was different).

### DIFF
--- a/lib/json-schema/validator.rb
+++ b/lib/json-schema/validator.rb
@@ -134,6 +134,9 @@ module JSON
       schema_uri = JSON::Util::URI.absolutize_ref(ref, parent_schema.uri)
       return true if self.class.schema_loaded?(schema_uri)
 
+      validator = self.class.validator_for_uri(schema_uri, false)
+      schema_uri = JSON::Util::URI.file_uri(validator.metaschema) if validator
+
       schema = @options[:schema_reader].read(schema_uri)
       self.class.add_schema(schema)
       build_schemas(schema)
--- a/resources/draft-03.json
+++ b/resources/draft-03.json
@@ -1,174 +1,174 @@
 {
-	"$schema" : "http://json-schema.org/draft-03/schema#",
-	"id" : "http://json-schema.org/draft-03/schema#",
-	"type" : "object",
-	
-	"properties" : {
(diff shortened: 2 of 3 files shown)


iainbeeston:
This sounds like a minor change but it's necessary in order to clear the schema cache between tests in the common test suite.

lib/json-schema/validator.rb
```diff
@@ -134,6 +134,13 @@ def load_ref_schema(parent_schema, ref)
       schema_uri = absolutize_ref_uri(ref, parent_schema.uri)
       return true if self.class.schema_loaded?(schema_uri)
 
+      begin
+        validator = self.class.validator_for_uri(schema_uri)
+        schema_uri = JSON::Util::URI.file_uri(validator.metaschema)
+      rescue JSON::Schema::SchemaError
```
iainbeeston: Yes... That is exactly right. I tried to avoid refactoring the other method. Perhaps it should take an option, so that it can return nil if it finds nothing? (Or, even better, I could replace validator_for with two methods - validator_for and validator_for! - and have the first method return nil and the second throw an exception?)

iainbeeston: I've already tried to clean up validator_for in #346 - perhaps I should take it further?

iainbeeston: Doing that would also help resolve #254

RST-J: I think having a bang and a non-bang method as you suggest is a good idea. And it should be easy to do as `validator_for_uri` itself explicitly checks for `nil` and if found raises the exception.

iainbeeston: I've corrected that now, by allowing `validator_for_uri` to take a second parameter that allows it to return `nil` (rather than raising an exception).

I chose that implementation (rather than bang methods) so that we don't change the default behaviour of any methods.

PR #367, opened 2016-12-04
Provide user invalid message

bhanubhakta:
This PR has created a flexibility for user to provide the invalidMessage from json-schema file itself, so that the error messages will be exactly what a user wants. Please review the PR and lets have discussions.


iainbeeston:
I'm sorry @bhanubhakta but this error message syntax isn't supported by the json schema format (see json-schema.org). Diverging from the spec would create all kinds of issues. I suggest that if you want to get it added to the json schema standard you raise the question on the json schema google group

bhanubhakta:
Thanks @iainbeeston

PR #368, opened 2016-12-25
Add Ruby 2.4.0 support.

marshall-lee:
Fixnum is deprecated in Ruby 2.4 so it should be replaced with Integer


lib/json-schema/attributes/allof.rb
```diff
@@ -29,7 +29,7 @@ def self.validate(current_schema, data, fragments, processor, validator, options
         end
 
         if !valid || !errors.empty?
-          message = "The property '#{build_fragment(fragments)}' of type #{data.class} did not match all of the required schemas"
+          message = "The property '#{build_fragment(fragments)}' of type #{type_of_data(data)} did not match all of the required schemas"
```
iainbeeston: Can you explain why this is necessary? (I assume something to do with the Fixnum/BigNum/Integer changes?)

lib/json-schema/util/array_set.rb
```diff
@@ -15,6 +15,6 @@ def include?(obj)
   private
 
   def convert_to_float_if_fixnum(value)
```
iainbeeston: It would be good to rename this method as well, as it still refers to fixnum

test/all_of_ref_schema_test.rb
```diff
@@ -21,15 +21,15 @@ def test_all_of_ref_subschema_errors
     errors = JSON::Validator.fully_validate(schema, data, :errors_as_objects => true)
     nested_errors = errors[0][:errors]
     assert_equal([:allof_0], nested_errors.keys, 'should have nested errors for each allOf subschema')
-    assert_match(/the property '#\/name' of type String did not match the following type: integer/i, nested_errors[:allof_0][0][:message])
+    assert_match(/the property '#\/name' of type string did not match the following type: integer/i, nested_errors[:allof_0][0][:message])
```
iainbeeston: Can we please remove these text changes, as they are not related to the change to ruby 2.4

test/min_items_test.rb
```diff
@@ -11,6 +11,6 @@ def test_minitems_nils
     errors = JSON::Validator.fully_validate(schema, [nil])
     assert_equal(errors.length, 1)
     assert(errors[0] !~ /minimum/)
-    assert(errors[0] =~ /NilClass/)
+    assert(errors[0] =~ /null/)
```
iainbeeston: Do you know why this has changed? (I assume something related to nil in ruby 2.4?)

lleger:
Tried to upgrade an app to Ruby 2.4 today and got bit by this. Any idea when you might be able to get a 2.4-compatible version pushed out?

iainbeeston:
It's the highest priority issue right now but unfortunately this PR needs some work before we can safely merge it. I can't give any timelines because it will depend on when myself of one of the other maintainers can spend some time on it. Sorry I can't be more precise but we are aware of the issue and are looking into it

iainbeeston:
I've split this into two separate commits and merged manually (one to remove references to Fixnum and one to make error messages consistently refer to json-schema types rather than ruby types), and I have released these changes under json-schema 2.8.0

Thanks for your contributions @marshall-lee

PR #377, opened 2017-03-03
Refactor common test suite

iainbeeston:
This pull request contains a number of improvements to how we use the common test suite:

1. Ignored tests from the common test suite are now stored in yaml file (rather than a large ruby hash)
1. We no longer abort running the tests if the common test suite cannot be updated (ie. when there is no internet connection)
1. Tests in the common test suite that are part of the core spec (ie. not optional) are now skipped rather than ignored
1. It ignores a number of new failing tests in draft6

Please see the commit messages for more details of each change

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -9,7 +9,14 @@ task :update_common_tests do
   unless File.read(".git/config").include?('submodule "test/test-suite"')
     sh "git submodule init"
   end
-  sh "git submodule update --remote --quiet"
+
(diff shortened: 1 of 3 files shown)


iainbeeston:
@RST-J Do you have time to code review this pull request? It should fix the failing tests in the common test suite and hopefully stop them failing unexpectedly in future

iainbeeston:
@RST-J thanks!

PR #382, opened 2017-05-17
improve handling of fragments

notEthan:
I've encountered a few problems dealing with fragments that this PR addresses:

- traversing the pointer through the document tries to instantiate the value as a json schema at every level, which is not valid when the value is an array, and not needed in any case. only the final value pointed to needs to be instantiated as a schema.
- when validate_schema is on, it validates the given json document, not the schema at the pointer. it should not assume that the given value is a json schema, only that the value at the pointer is.
- strings containing escaped values, either URI escaped in the fragment, or escaped with `~0` / `~1` as defined in https://tools.ietf.org/html/rfc6901 are not handled correctly.

I made a class JSON::Schema::Pointer to abstract the parsing and evaluation of pointers, and updated other code around this.


lib/json-schema/schema.rb
```diff
@@ -6,6 +6,8 @@ class Schema
     attr_accessor :schema, :uri, :validator
 
     def initialize(schema,uri,parent_validator=nil)
+      raise ArgumentError, "schema must be a hash; got: #{schema.inspect}" unless schema.is_a?(Hash)
```
notEthan: this change is kind of unrelated, but it results in a better error message when an array is passed as the `schema` argument (as the previous fragment-handling code was prone to do) - before the error message would end up being `TypeError: no implicit conversion of String into Integer` when trying to call `@schema['id']` below, which is rather obscure.

lib/json-schema/validator.rb
```diff
@@ -55,6 +56,11 @@ def initialize(schema_data, data, opts={})
       @data = initialize_data(data)
       @@mutex.synchronize { build_schemas(@base_schema) }
 
+      # If the :fragment option is set, try and validate against the fragment
+      if opts[:fragment]
+        @base_schema = schema_from_fragment(@base_schema, opts[:fragment])
+      end
```
notEthan: this is moved before the validate_schema block so that the correct schema, at the fragment pointer, is validated

lib/json-schema/pointer.rb
```diff
@@ -0,0 +1,124 @@
...
+      # #/foo%20bar
+      #
+      # => ['foo bar']
+      def self.parse_fragment(fragment)
+        fragment = Addressable::URI.unescape(fragment)
+        match = fragment.match(/\A#/)
+        if match
+          parse_pointer(match.post_match)
+        else
+          raise(PointerSyntaxError, "Invalid fragment syntax in #{fragment.inspect}: fragment must begin with #")
```
notEthan: I've also tried to improve the information given in these error messages, as the previous `"Invalid fragment resolution for :fragment option"` was not very useful to debug

lib/json-schema/validator.rb
```diff
@@ -64,42 +70,14 @@ def initialize(schema_data, data, opts={})
...
-
-      # If the :fragment option is set, try and validate against the fragment
-      if opts[:fragment]
-        @base_schema = schema_from_fragment(@base_schema, opts[:fragment])
-      end
     end
 
     def schema_from_fragment(base_schema, fragment)
       schema_uri = base_schema.uri
-      fragments = fragment.split("/")
```
notEthan: another fix I just noticed: this should be `split("/", -1)`; as is empty string keys would be mishandled. `#/foo//` should evaluate to fragments `['foo', '', '']` and in json `{"foo": {"": {"": 7}}}` should evaluate to `7`.

notEthan:
added https://github.com/ruby-json-schema/json-schema/pull/383 and rebased this against it

notEthan:
this thing alive?

iainbeeston:
@notEthan Sorry, I've been immensely busy at work and at home for several weeks. I'll take a proper look at this once things calm down here

notEthan:
sounds good, thanks

lib/json-schema/schema.rb
iainbeeston: I'm trying (slowly) to move the codebase away from explicit type checking towards duck typing. Could you please alter this check from `schema.is_a?(Hash)` to `schema.respond_to?(:to_hash)`, and the line below from `@schema = schema` to `@schema = schema.to_hash`?

lib/json-schema/pointer.rb
```diff
@@ -0,0 +1,124 @@
...
+      # initializes a JSON::Schema::Pointer from the given representation.
+      #
+      # type may be one of:
+      #
+      # - :fragment - the representation is a fragment containing a pointer (starting with #)
+      # - :pointer - the representation is a pointer (starting with /)
+      # - :reference_tokens - the representation is an array of tokens referencing a path in a document
+      def initialize(type, representation)
+        @type = type
+        if type == :reference_tokens
```
iainbeeston: I can't see anywhere where this is used. Is it needed? (And if not we can remove the last clause in `representation_s` too)

lib/json-schema/pointer.rb
```diff
@@ -0,0 +1,124 @@
...
+          end
+          value[token]
+        end
+      end
+
+      def to_s
+        "#<#{self.class.name} #{@type} = #{representation_s}>"
+      end
+
+      def representation_s
```
iainbeeston: Could we make this and the methods below private, to remove them from the public api of the class?

lib/json-schema/pointer.rb
```diff
@@ -0,0 +1,124 @@
...
+      end
+
+      # initializes a JSON::Schema::Pointer from the given representation.
+      #
+      # type may be one of:
+      #
+      # - :fragment - the representation is a fragment containing a pointer (starting with #)
+      # - :pointer - the representation is a pointer (starting with /)
+      # - :reference_tokens - the representation is an array of tokens referencing a path in a document
+      def initialize(type, representation)
```
iainbeeston: Actually, the more I think about this the more I feel that this aught to be refactored from a single Pointer class, into a Pointer parent class and a subclass for each type (eg. FragmentPointer and AbsolutePointer maybe?). Right now most instance methods work differently based on the type, and it feels like an instance of the [Switch Statements Smell](http://wiki.c2.com/?SwitchStatementsSmell), where polymorphism would work better.

iainbeeston:
You might need to rebase this, as I've merged a fragment-related bugfix (#217)

notEthan:
oh my, what a long time I have neglected this. back to it. rebased and addressed some comments.

lib/json-schema/pointer.rb
notEthan: I used it in code that I have been working on that relies on this Pointer class - I use the array of reference tokens there. dealing with the string (pointer or fragment) in code is less useful than the tokens. I'd like this to be part of the API of this class.

lib/json-schema/pointer.rb
notEthan: I've made this method private but #pointer and #fragment should be public. I've added method documents for those.

lib/json-schema/pointer.rb
notEthan: I started down this path but on further reflection I think I disagree. the class represents the same sort of pointer no matter what the external representation it came from. the only method that is any different, once the external representation has been parsed, is #to_s and that is only to keep track of where it came from - it's not really a part of what the Pointer is, just something I found useful for debugging.

notEthan:
if you agree that a Pointer is a Pointer, I hope this is ready to be merged. if you feel strongly that it should be polymorphic, I can do that.

notEthan:
hmm. test failures on jruby that I'm not getting locally. one looks related to this change, others don't. will address what I can.

notEthan:
I didn't take the time to actually grok the caching test failures but the last commit here seems to fix 'em

notEthan:
I put a commit fixing a require that was missing, causing failures on jruby, ahead of the content of this PR, it can be merged together or separately

notEthan:
ping on this, hopefully you have time to look at it and we can bring it home

skryukov:
Hi @iainbeeston thanks for the great gem! Could you please look at this PR again? Keys with slashes are all over the place in the swagger specs =)

scalp42:
@iainbeeston is it still maintened? Would it be possible to add more contributors as a lot of PRs are sitting? Thanks!

iainbeeston:
@scalp42 I'm afraid that the gem does *not* have any active maintainers at the moment. If you're interested in becoming one please comment on #423

PR #383, opened 2017-05-17
webmock 3 drops ruby 1.9 support; specify this in its gemfile

notEthan:


iainbeeston:
Good catch, thanks for the contribution

PR #386, opened 2017-06-30
Data that looks like a URI but is invalid should be treated as a string

iainbeeston:
There's a bug where if you pass a string as data and it looks like a
URI, we try to parse it and load the data from that location. However,
if parsing fails, we should give up trying to load remote data and
treat the data as a string instead.


iainbeeston:
@RST-J Could I please get a review of this too?

RST-J:
Sorry, I'm currently not able to review. I have RL stuff to deal with.

⁣Bis neulich
Jonas

Gesendet mit BlueMail ​

iainbeeston:
@RST-J No problem, I understand completely (sorry for spamming you with requests!)

ekohl:
There's at least a merge conflict so it needs a rebase.

PR #387, opened 2017-07-05
Always download the common test suite before running tests

iainbeeston:
In #377 I made it possible to run the tests without having internet
access. But it seems that in the process I also disabled the code that
downloads the common test suite when you first check out the
codebase. That's bad, but what makes it worse is that travis hasn't
been running the common test suite because of it.

This changes the behaviour so that the tests download the common test
suite (but does not update it) before running. If you don't have
internet access it continues without it. To update the common test
suite to the latest version, you now need to run
`UPDATE_TEST_SUITE=true rake download_common_tests`.


iainbeeston:
@RST-J Could I please get a quick review of this? It's a change to the test suite rather than user-facing code.

iainbeeston:
Ok this isn't fixing an issue on travis as I forst thought. It seems that by default travis performs a gut submodule update after checkout, which is what we're doing here. However, it would be helpful if we did this automatically for contributors too

PR #405, opened 2018-02-15
strict fixes, needs tests

kwstannard:


lib/json-schema/schema/validator.rb
```diff
@@ -20,7 +20,12 @@ def extend_schema_definition(schema_uri)
       end
 
       def validate(current_schema, data, fragments, processor, options = {})
-        current_schema.schema.each do |attr_name,attribute|
+        schema = current_schema.schema
+        if options[:strict] && schema['required'].nil? && schema['properties']
+          schema['required'] = schema['properties'].keys
```
kwstannard: If it isn't obvious what is happening here, I changed strict mode to automatically set "required" to every property key if no required is already defined.

kwstannard:
Opened a pull on the wrong remote, sorry. Feel free to use this.

PR #406, opened 2018-02-27
Allow override of required in strict mode

kwstannard:
Current functionality: required field is ignored in strict mode in favor
of making all properties required.

Proposed functionality: required field is used to override the strict
mode.

Why: This will allow people to save time by using strict mode even if some
properties are optional.


pboling:
Without looking at the documentation this is how I would have assumed it would work, so I think this is a good change.

kwstannard:
Clearing out my pull requests that are never going to get merged.

PR #413, opened 2018-09-11
Add support for OAS3

myxaluch:
Hello!

In our company we try to use JSON scheme for Swagger documentation also for spec's response check, via `json-scema`, and faced with one problem: after upgrade scheme to Open Api v3, we can't use it completely, because JSON scheme and OAS become more different than before. So, I propose to add support for OAS3 in `json-scheme` gem, for this case, and, first of all, add support for `nullable` option in property description.


lib/json-schema/attributes/type_oas_attribute.rb
```diff
@@ -0,0 +1,13 @@
+require 'json-schema/attributes/type_v4'
+
+module JSON
+  class Schema
+    class TypeOasAttrbiute < TypeV4Attribute
```
jbmeerkat: typo, `Attrbiute` instead of `Attribute`

myxaluch:
We've decided to add support via own library - https://github.com/myxaluch/json-schema-oas-support, so I'll close this PR, sorry for inconvenience

PR #420, opened 2018-10-14
Add :property_fragment to errors hash, to show which required property was missing

ndbroadbent:
I understand the reasoning behind `:fragment => "#/"` for a required property that is missing from the root object. But I really need to know *which* required property is missing, and I didn't want to parse it out of the message string.

I've added a new `:property_fragment` to the validation error hash, and this is a more consistent way of detecting which field is causing the validation error. For required fields, it will show the property that is missing. And it will be the same for fields that have the wrong data type, etc (because it falls back to `:fragment` by default.)

This change should be fully backwards compatible, and it just adds a new key to the error hash.

### DIFF
--- a/lib/json-schema/attribute.rb
+++ b/lib/json-schema/attribute.rb
@@ -10,8 +10,8 @@ module JSON
         "#/#{fragments.join('/')}"
       end
 
-      def self.validation_error(processor, message, fragments, current_schema, failed_attribute, record_errors)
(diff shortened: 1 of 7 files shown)


RST-J:
LGTM

ssrihari:
plz2merge @iainbeeston

pboling:

PR #425, opened 2018-11-09
Add const validator to draft6.

torce:
Ignored all draft7 tests in common suite.


pboling:

torce:
@bastelfreak Since this feature was already merged in #471, I have reduced this PR to just enable the "const" tests of the common test suite which was the only missing part of #471.

bastelfreak:
@torce thanks for the work!

Issue #423, opened 2018-10-16
The future of the JSON schema gem


iainbeeston:
@ruby-json-schema/owners 

I have a lot of ideas for refactoring, modernising and slimlining the json-schema gem, as well as adding support for draft 6, 7 and eventually 8. However, I know everyone is busy and reviewing pull requests for json-schema isn't always a high priority.

I'm concerned that the project has become neglected, and with that in mind, I feel that someone needs to step up to take a bit more of a lead to steer the project into a better place. I'd be happy to be that person, and would be prepared to take on the responsibility of making sure the project is well-maintained in the future.

Could you please let me know your thoughts on the idea of me taking the lead on the json-schema project, which would entail cleaning it up and refactoring it as I see fit, and merging pull requests without the agreement of another maintainer. If nobody objects, I'll assume that everyone is busy with other things and is happy for me to push the project forward alone.

Finally, if you're no longer interested in maintaining the project and don't wanted to be "pinged" by contributors on issues or pull requests, let me know and I can change you from being listed as an "Owner" to a "Contributor".

RST-J:
I'm fine with you taking the lead as proposed. Although I probably do not have much time to put into the project (and have been absent for about 1.5 to 2 years for personal reasons), I'd try to help with reviews if possible. Maybe also get back into development but I don't want to promise something I cannot keep.

we4tech:
Hi @iainbeeston Let me know if you need any help.

iainbeeston:
I've been thinking about this more since I opened the issue, and I've realise that I was being unrealistic. I have a lot of other projects competing for my limited time, on top of work and family life. I would love to bring this project up to date with the latest json schema drafts and refactor it into something more maintainable and see it progress into the future, but I can't spare enough time to make that happen. 

Perhaps instead this thread should start the search for new maintainers?

iainbeeston:
Is anyone interested in becoming a maintainer of json-schema?

There are a lot of open pull requests that need merging, but it's important to understand how they fit into the existing codebase before merging, as they might introduce new issues (the existing test suite is *not* comprehensive for more complex schema, such as loading from uris or fragments - see #248). Also the gem supports a number of unofficial extensions to the specification (such as default values) that add complexity.

[...]

PR #431, opened 2019-02-18
Support Draft 7

Martouta:
Support Draft 7. Not all of it but it means to support what we need for https://github.com/3scale/porta/pull/593

### DIFF
--- a/VERSION.yml
+++ b/VERSION.yml
@@ -1,3 +1,3 @@
-major: 2
-minor: 8
-patch: 1
+major: 3
+minor: 0
+patch: 0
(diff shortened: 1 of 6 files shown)


heynan0:
I think someone could build upon your work here to support more drafts.

Martouta:
@heynan0 you should check https://github.com/voxpupuli/json-schema/issues/423 and https://github.com/davishmcclurg/json_schemer

Martouta:
btw, that info may also be useful for @secretsurf @josemigallas (assuming that you are still part of the 3scale system/porta team) ☝🏼

TL;DR: This repo/gem is not really thaaaat maintained anymore (kinda looking for maintainers) and does not support new draft versions but the other repo/gem does.

bastelfreak:
Hi, we are happy to review new patches for the project and would also accept new maintainers :)

PR #434, opened 2019-04-15
Bugfix: Fix fragment when used with extended schema

zdennis:
# What this PR does

This PR fork resolves an issue with extended schema validation.

From the original author of the patch (circa 2016):

> fix fragment when used with extended schema
> 
> This refactors some of `Validator` to fix an issue with validating a fragment of an extended schema against a json object. The problem is that while navigating through the schema, we are using the configured `:version` option as the validator for each partial schema. This is an issue because if the root of the schema declares a `$schema`, that information is lost in the traversal.
> 
> The refactor does two things. Firstly, it does not write back into `@options[:version]` and it stops using it outside of the constructor entirely. The only place that should know about this is the initialized schema, and from that point on, the canonical validator should be the schema object's validator since it makes a decision about which to use in its constructor. Now we can reduce complexity by using that validator in the `:validate_schema` option branch.
> 
> The other place that used `:version` is the `schema_from_fragment` method, which is where the bug exists. What we're doing here, instead of dealing with schemas *and* hashes during the iterator, is pulling the schema object out, digging through it using the fragments, and then wrapping the result in a schema using the `base_schema`'s validator.
>
> Tests have been included that demonstrate the problem.

### DIFF
--- a/lib/json-schema/validator.rb
+++ b/lib/json-schema/validator.rb
@@ -41,8 +41,7 @@ module JSON
       @options = @@default_opts.clone.merge(opts)
       @errors = []
 
-      validator = self.class.validator_for_name(@options[:version])
-      @options[:version] = validator
+      configured_validator = self.class.validator_for_name(@options[:version])
       @options[:schema_reader] ||= self.class.schema_reader
 
       @validation_options = @options[:record_errors] ? {:record_errors => true} : {}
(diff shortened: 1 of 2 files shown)


zdennis:
We weren't sure if updating the common test-suite made sense so we're happy to take that commit off.

pboling:

bastelfreak:
thanks for the PR. I merged it in https://github.com/voxpupuli/json-schema/pull/464

Issue #404, opened 2018-02-11
"additionalProperties: false" is ignored if type specified in array


elasti-ron:
Given the object
```json
object = {
  "a_string": "this property is allowed by the schema",
  "an_extra_string": "the schema does not allow this property"
}
```

The following schema should raise a validation error but doen't: 
```json
// should raise error but doesn't
{
  "type": [ "object", "null" ],
  "properties": {
    "a_string": {
      "type": "string"
    }
  },
  "additionalProperties": false
}
```

This occurs even if only a single type is specified in the array of type options:
```json
// should raise error but doesn't
{
  "type": [ "object" ],
  "properties": {
    "a_string": {
      "type": "string"
    }
  },
  "additionalProperties": false
}
```

Note that this does not occur when the type is specified directly and not in an array.  In other words, the following schema does raise the expected error:

```json
// raises error as expected
{
  "type": "object",
  "properties": {
    "a_string": {
      "type": "string"
    }
  },
  "additionalProperties": false
}
```

This issue may be related to #369

PR #438, opened 2019-08-17
AdditionalProperties not being applied when value could be array

dharanikumar-19:
This fix should will resolve this issue #404 

Given the following schema:

`// should raise error but doesn't
{
  "type": [ "object", "null" ],
  "properties": {
    "a_string": {
      "type": "string"
    }
  },
  "additionalProperties": false
}
 `
The sample data
`object = {
  "a_string": "this property is allowed by the schema",
  "an_extra_string": "the schema does not allow this property"
}`

Previously it does not raise any error

After the fix
Following error will throw
`JSON::Schema::ValidationError: The property '#/' contains additional properties ["an_array"] outside of the schema when none are allowed`


pboling:

bastelfreak:
Hi @dharanikumar-19, thanks for this PR. Are you able to rebase and add tests for the change?

PR #442, opened 2020-02-02
Always return duplicates from JSON::Util::URI.parse

jhass:
Some callsides seem to modify the returned URI instance, messing up
the cache. This can lead to the wrong schema fragment being used
when multiple schemas with overlapping fragment paths are loaded
and validated against.

Unfortunately I didn't manage to construct an integrative test case
and I'm not sure how much worth something would be that just checks this
method is indeed returning duplicates in any case

### DIFF
--- a/lib/json-schema/util/uri.rb
+++ b/lib/json-schema/util/uri.rb
@@ -65,12 +65,8 @@ module JSON
           return uri.dup
         else
           @parse_cache ||= {}
-          parsed_uri = @parse_cache[uri]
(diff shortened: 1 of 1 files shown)


pboling:

bastelfreak:
Hi @jhass ,
thanks for bringing this up. The project was recently migrated to Vox Pupuli. Are you able to rebase and add a test now?

jhass:
No, after two years I longer have the original context.

PR #445, opened 2020-10-13
Force json version to 2.2.0 for ruby lower than 2.0

h0lyalg0rithm:
Json 2.3.0 uses some new features which are only present in ruby 2.0 and above.
I have updated the Gemfile to prevent users of ruby 1.9 to install the newer json library and run in to runtime issues

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,5 +2,5 @@ source "https://rubygems.org"
 
 gemspec
 
-gem "json", ">= 1.7", :platforms => :mri_19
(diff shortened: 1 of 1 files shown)


h0lyalg0rithm:
@Vasfed What do you think about this?

h0lyalg0rithm:
@iainbeeston @kylog @jarib

Vasfed:
Ruby 1.9 EOL was in 2015, I think we should drop support for it

pboling:

bastelfreak:
Hi,
I'm going to close this PR. The master branch currently only supports Ruby 2.5 and newer.

Issue #449, opened 2021-01-06
URI.open via Kernel#open is deprecated


estolfo:
Hi!
Ruby 3.0 has removed support for `URI.open` via `Kernel#open`, making these lines fail:

- [reader.rb#L121](https://github.com/ruby-json-schema/json-schema/blob/v2.8.1/lib/json-schema/schema/reader.rb#L121)
- [validator.rb#L596](https://github.com/ruby-json-schema/json-schema/blob/v2.8.1/lib/json-schema/validator.rb#L596)

I was going to open a PR to change those lines but I noticed in your CI that you run your tests against ruby versions back to 1.9.
It seems that `URI.open` was made public in ruby 2.4. I guess the two options would be to drop support for ruby < 2.4 or use `URI.send(:open, ...)`, which should work for all versions of ruby that you support (I have not verified this yet).

I'm happy to open a pull request and update your test matrix, just let me know what you'd like to do!
Thanks!

PR #451, opened 2021-02-11
Support Ruby 3

vivekbisen:
#449


pboling:

PR #456, opened 2021-09-02
update README.md/gemspec;  migrate to GitHub actions

bastelfreak:


TuningYourCode (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'voxpupuli'
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Ruby 3.0
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.0'
+        env:
+          BUNDLE_WITHOUT: release
```
ekohl, line 19: This is useless. There is no bundler involved so it can be dropped.

.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
...
+        run: gem push *.gem
+        env:
+          GEM_HOST_API_KEY: '${{ secrets.RUBYGEMS_AUTH_TOKEN }}'
+      - name: Setup GitHub packages access
+        run: |
+          mkdir -p ~/.gem
+          echo ":github: Bearer ${{ secrets.GITHUB_TOKEN }}" >> ~/.gem/credentials
+          chmod 0600 ~/.gem/credentials
+      - name: Publish gem to GitHub packages
+        run: gem push --key github --host https://rubygems.pkg.github.com/voxpupuli *.gem
```
ekohl, line 32: Please don't hardcode voxpupuli. See https://github.com/voxpupuli/.github/blob/master/workflow-templates/gem-release.yml for an example.

PR #437, opened 2019-05-27
Ruby 2.6 support

lukeasrodgers:
Addresses an `IPAddr` validation error issue with ruby 2.6.
Fixes #430

PR #457, opened 2021-09-02
Add Ruby 2.6/2.7 support

bastelfreak:
This is a resubmission of https://github.com/voxpupuli/json-schema/pull/437

### DIFF
--- a/lib/json-schema/attributes/formats/ip.rb
+++ b/lib/json-schema/attributes/formats/ip.rb
@@ -11,7 +11,7 @@ module JSON
         begin
           ip = IPAddr.new(data)
         rescue ArgumentError => e
-          raise e unless e.message == 'invalid address'
+          raise e unless e.message.start_with?('invalid address')
         end
 
         family = ip_version == 6 ? Socket::AF_INET6 : Socket::AF_INET
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,6 +15,8 @@ jobs:
       matrix:
         include:
           - ruby: "2.5"
+          - ruby: "2.6"
+          - ruby: "2.7"
             coverage: "yes"
     env:
       COVERAGE: ${{ matrix.coverage }}


lib/json-schema/attributes/formats/ip.rb
```diff
@@ -11,7 +11,7 @@ def self.validate(current_schema, data, fragments, processor, validator, options
         begin
           ip = IPAddr.new(data)
         rescue ArgumentError => e
-          raise e unless e.message == 'invalid address'
+          raise e unless e.message.start_with?('invalid address')
         end
```
ekohl, line 15: I feel that parsing exceptions is fragile and I'd implement it like this (roughly):
```ruby
class IPFormat < FormatAttribute
  def self.validate(current_schema, data, fragments, processor, validator, options = {})
    return unless data.is_a?(String)

    unless self.regex.match?(data)
      error_message = ....
      validation_error(...)
    end
  end

  def self.regex
    raise NotImplementedError
  end
end

class IP4Format < IPFormat
  def self.ip_version
    4
  end

  def self.regex
    Resolv::IPv4::Regex
  end
end
```

I do think `regex.match?()` may be Ruby 2.5 only but you can use `=~` for older Rubies.

However, I'm not sure if this regex is available in older Rubies.

os6sense:
Any plans to deal with the deprecation warnings that are raised when using 2.7.x?  e.g. 

```
/home/runner/work/M3/M3/vendor/bundle/ruby/2.7.0/gems/json-schema-2.8.1/lib/json-schema/schema/reader.rb:121: warning: calling URI.open via Kernel#open is deprecated, call URI.open directly or use URI#open
```

bastelfreak:
@os6sense we do plan to fiy those in the future when we've some time. I also havent reviewed the remaining PRs yet. Feel free to take a look as well or zo provide a new PR that fixes this. every contribution is welcome!

bastelfreak:
@ekohl are you fine with merging this for now and update it later? Or can you submit your code as a PR?

ghoneycutt (approved):

PR #458, opened 2021-09-02
cleanup CI environment

bastelfreak:


.github/workflows/test.yml
```diff
@@ -4,9 +4,6 @@ on:
   - pull_request
   - push
 
-env:
-  BUNDLE_WITHOUT: release
-
```
ekohl, line 9: This can stay, but you should clean it up in the release workflow.

bastelfreak: done

PR #462, opened 2021-11-27
Call URI.open directly / Drop Ruby 2.4 support

bastelfreak:
Removes warning: "calling URI.open via Kernel#open is deprecated, call URI.open directly or use URI#open"

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - ruby: "2.4"
           - ruby: "2.5"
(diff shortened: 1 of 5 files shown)


bastelfreak:
Rebase of https://github.com/voxpupuli/json-schema/pull/459

bastelfreak:
comment from the second commit:
```
The codechange is required to work properly on newer Ruby versions, but
breaks Ruby 2.4. Since Ruby 2.4 is EoL since a long time it's fine to
drop support for it. This also makes it easier to add Ruby 3 support
later.
````

sebastianrakel (approved):

PR #465, opened 2022-02-16
Cache schema and re-use it to speed up

ganmacs:
In my environment,  I call `JSON::Validator.validate` many times with huge JSON Schema definitions and it takes a lot of time to return value. After some investigations, I found that `JSON::Validator#initialize_schema` in `JSON::Validator#initialize` takes almost all time.

This patch removes `@data` from instance variables of JSON::Validator so that it can avoid a lot of duplicated object creation. it's really useful and efficient when you want to apply the same schema but different data multiple time.

```
schema = ....
validator = JSON::Validator.new(schema)

validator.validate(data1)
validator.validate(data2)
validator.validate(data3)
```

## flamegraph

Source code
https://gist.github.com/ganmacs/29f6217bbd3658d7f1f0010d0cadedd7

```
$ bundle exec ruby test.rb
$ stackprof --flamegraph stackprof.dump > flamegraph
$ stackprof --flamegraph-viewer=flamegraph
```

Before
Screen Shot 2022-02-16 at 14 23 11

After
Screen Shot 2022-02-16 at 14 25 33


ghoneycutt (approved):

bastelfreak:
@ganmacs thanks for the PR. Could you rebase please to pull in our latest CI changes?

smortex (approved):
These changes make sense, loading the schema and the data at the same time does not really makes sense so changing the API to split these two operations is I think a good move.

ekohl (approved):
I missed this, but it looks good to me.

PR #471, opened 2022-03-15
Add const attribute support 

jeremie-stripe:
Added in draft 6. See [documentation](https://json-schema.org/understanding-json-schema/reference/generic.html#constant-values) on json-schema.org

### DIFF
--- a/test/draft6_test.rb
+++ b/test/draft6_test.rb
@@ -1,0 +1,24 @@
+# encoding: utf-8
+require File.expand_path('../support/test_helper', __FILE__)
+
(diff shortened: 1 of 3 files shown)


test/draft6_test.rb
```diff
@@ -0,0 +1,24 @@
...
+      }
+    }
+
+    data = {:a => "foo", :b => 6}
+    assert_valid schema, data
+
+    data = {:a => 6, :b => "foo"}
+    refute_valid schema, data
+  end
+end
```
bastelfreak, line 24: please add the missing \n in line 24

PR #478, opened 2023-02-01
Switch to latest voxpupuli-rubocop

bastelfreak:
Both ruby version are EoL. We already merged a breaking change to master, so I think it's good to drop those old ruby versions.


ekohl:
metadata-json-lint pulls this in, so it can end up being painful for a lot of users. IIRC bundler in Ruby 2.5 isn't able to deal well with the required ruby version. Take that into consideration.

bastelfreak:
@ekohl metadata-json-lint uses json-schema `'>= 2.8', '< 4.0'`. The next major release will be 4.0 so metadata-json-lint would not automatically pull it in. I suggest we bump Ruby here and when we bump it in metadata-json-lint to 2.7 we can pull in metadata-json-lint 4.x.

bastelfreak:
We now removed Ruby 2.6 and older in another PR. This one just updates rubocop. I left a bit of stuff in the todo file, otherwise the PR would have been too huge. I will clean that up in a followup PR.

Gemfile
```diff
@@ -10,12 +10,4 @@ end
...
-end
-
-group :tests do
-  gem 'rubocop', '~> 1.12.0' # newer version require Ruby 2.6
-  gem 'rubocop-minitest', '~> 0.19.0' # newer version requires Ruby 2.6
-  gem 'rubocop-performance', '~> 1.13.0' # newer version requires Ruby 2.6
-  gem 'rubocop-rake', '~> 0.6.0' # latest
-  gem 'rubocop-rspec', '~> 2.4.0' # newer version require Ruby 2.6 or rubocop 1.19
-end
+  end
```
ekohl, line 13: Looks misaligned
```suggestion
end
```

.rubocop_todo.yml
```diff
@@ -1,93 +1,107 @@
...
 # versions of RuboCop, may require this file to be generated again.
 
 # Offense count: 1
-# Cop supports --auto-correct.
+# This cop supports safe autocorrection (--autocorrect).
+# Configuration parameters: EnforcedStyleAlignWith.
+# SupportedStylesAlignWith: either, start_of_block, start_of_line
+Layout/BlockAlignment:
+  Exclude:
+    - 'Gemfile'
```
ekohl, line 15: This is what I highlighted so please regenerate after

bastelfreak: done

Gemfile
bastelfreak: fixed

PR #481, opened 2023-02-17
CI: Only run on pull requests & Use latest GitHub workflows

bastelfreak:


sebastianrakel (approved):

.github/workflows/test.yml
```diff
@@ -2,7 +2,6 @@ name: Test
 
 on:
   - pull_request
-  - push
```
ekohl, line 5: Note that if you do it on push (perhaps we can even modify it to only the default branch) you get a cache for PRs on the Ruby build. Without it the cache is scoped to each individual, since reusing another user's cache could lead to security issues.

bastelfreak: I updated it on https://github.com/voxpupuli/json-schema/pull/486

PR #484, opened 2023-02-17
Fix more rubocop violations

bastelfreak:


.rubocop.yml
```diff
@@ -8,3 +8,15 @@ require:
 AllCops:
   TargetRubyVersion: 2.5
   NewCops: enable
+
+Style/TrailingCommaInHashLiteral:
```
bastelfreak, line 12: we agreed on using those in https://github.com/voxpupuli/voxpupuli-test/pull/86

arjenz (approved):

PR #490, opened 2023-02-17
fix more rubocop violations

bastelfreak:


b4ldr:
LGTM and seems like just style changes but not too familiar with this repo

Issue #475, opened 2022-11-18
Type defined as string gets rejects a string integer


andykoch:
I would expect a string that is just a number to pass validation as a string.

 JSON::Validator.fully_validate({type: "string"}, "123")
=> ["The property '#/' of type integer did not match the following type: string in schema 42b11e45-3593-5a7f-9e6a-a0f0f508e253"]

 JSON::Validator.fully_validate({type: "string"}, "123-Foo")
=> []

When I use the same schema and value in a third party validator, it passes fine.

andykoch:
I dug into the source and discovered the string was getting JSON.parse() called on it, which returns integer 123. To address that I appended a to_json.

`JSON::Validator.fully_validate({type: "string"}, "123".to_json)`

`"123".to_json` returns `"\"123\""`

PR #497, opened 2023-07-21
Integer as string should be defined as string

mohanapriya2308:
Fix for #475 

`parse_data` parses integer string also to integers. This breaks when integer string is expected to be string. To fix this, introduced new param `parse_integer` when set to false, integers are not parsed.
 
_Before_
 JSON::Validator.fully_validate({type: "string"}, "123")
=> ["The property '#/' of type integer did not match the following type: string in schema"]

_After_
 JSON::Validator.fully_validate({type: "string"}, "123", {"parse_integer" => false)
=> []


bastelfreak:
@mohanapriya2308 thanks for the fix!

PR #507, opened 2024-01-05
Support frozen string literals

mostlyobvious:


.github/workflows/test.yml
```diff
@@ -17,24 +17,35 @@ jobs:
...
+            rubygems: default
           - ruby: "3.0"
+            rubygems: default
           - ruby: "3.1"
+            rubygems: default
             coverage: "yes"
           - ruby: "3.2"
+            rubygems: default
     env:
+      RUBYOPT: --enable-frozen-string-literal
```
ekohl, line 40: Would it be better to add the frozen string literal comment to the file(s)? That way Ruby can use frozen strings even at runtime.

ekohl (commented):
I'd prefer to separate out the CI changes. It's probably better to drop Ruby 2.5 (and 2.6). I'd keep 2.7 for a bit longer, even though it's already EOL.

The actual change to add `+` to the string is correct since it needs to be mutable.

.github/workflows/test.yml
mostlyobvious: It would be better for the consumers of the gem. It changes nothing for checking the compatibility of running with strings frozen.

I'd prefer to separate it out.

mostlyobvious:
> I'd prefer to separate out the CI changes. It's probably better to drop Ruby 2.5 (and 2.6). I'd keep 2.7 for a bit longer, even though it's already EOL.

I'd prefer to not make any CI changes either. They were unfortunately necessary. While I share the sentiment to drop older EOL Ruby versions, I do not see much recent activity around it [https://github.com/voxpupuli/json-schema/pull/478].

I also do not see how those Ruby CI changes you eventually want are blocking the improvement of running with frozen string literals enabled. There is no conflict in making Ruby 2.5 matrix a bit better with updated-and-fixed Rubygems, even if it is going to be dropped soon. And it is something already done.

PR #508, opened 2024-03-15
Ensure compatibility with `ruby --enable-frozen-string-literal`

casperisfine:
Ref: https://bugs.ruby-lang.org/issues/20205

`json-schema` is used by the `rbs` gem, which is tested as part of the `ruby-core` CI, and it's failing with frozen string literals are enabled by default.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,9 @@ jobs:
           - ruby: "3.1"
             coverage: "yes"
           - ruby: "3.2"
+          - ruby: "3.3"
+          - ruby: "3.3"
+            rubyopt: "--enable-frozen-string-literal"
     env:
       COVERAGE: ${{ matrix.coverage }}
     steps:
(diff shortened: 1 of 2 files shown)


casperisfine:
Thanks for the extremely timely merge :)

casperisfine:
As a maintainer I know it's annoying to be asked for a release, but in this case I arrived here because I'm trying to get `rbs` test suite to pass with `--enable-frozen-string-literal` and it depend on `json-schema`. I can use a `git` gem for now, but not sure they'll be willing to merge my fixes without a proper release cut.

So I'd owe you one if you could cut a release soon 🙏

bastelfreak:
already on it :) 
https://github.com/voxpupuli/json-schema/pull/509

casperisfine:
😻

mostlyobvious:
considering https://github.com/voxpupuli/json-schema/pull/507 resolved too

PR #512, opened 2024-05-22
Add missing require 'stringio'

jmarrec:
In some cases, such a statically built ruby 3.x, I definitely get `NameError: uninitialized constant IO::StringIO`

### DIFF
--- a/lib/json-schema/validator.rb
+++ b/lib/json-schema/validator.rb
@@ -5,6 +5,7 @@ require 'digest/sha1'
 require 'date'
 require 'thread'
 require 'timeout'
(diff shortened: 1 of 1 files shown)


bastelfreak:
I think that's a bug in the ruby version you're using, but I think it won't hurt us to explicitly require it. I had similar issues with `json` in other projects.

jmarrec:
@bastelfreak What I meant by "statically built ruby" iss that I use a ruby built with --enable-static and --with-static-linked-ext which is indeed unusual.
Thanks for the merge!

Issue #514, opened 2024-07-05
Reference resolution issue


bolshakov:
We use `json-schema` in a multi-threaded environment, and as outlined in issues #310 and #242, the heavy use of caching at the class level could potentially lead to race conditions.

We encounter this problem intermittently. In trying to reproduce the issue, I discovered a problem with the `JSON::Util::URI.normalize_ref` method, which can be reproduced even without using multiple threads:

```ruby
assert_equal Addressable::URI.parse('http://www.example.com/#foo'), JSON::Util::URI.normalize_ref("#foo", 'http://www.example.com')
assert_equal Addressable::URI.parse('http://www.example.net/#foo'), JSON::Util::URI.normalize_ref("#foo", 'http://www.example.net')
```

The test above fails with the following output:

```ruby
Minitest::Assertion: --- expected
+++ actual
@@ -1 +1 @@
-#<Addressable::URI:0xb54 URI:http://www.example.net/#foo>
+#<Addressable::URI:0xb68 URI:http://www.example.com#foo>
```

The errors disappear if the cache is cleared between assertions. Apparently, the implementation of the `.normalize_ref` method stores parsed `#foo` in the cache and subsequently modifies it by extending it with the base URL. Therefore, the same `ref` could be resolved differently, ignoring the `base` argument entirely.

### Steps to Reproduce

1. Run the following test:
   ```ruby
   assert_equal Addressable::URI.parse('http://www.example.com/#foo'), JSON::Util::URI.normalize_ref("#foo", 'http://www.example.com')
   assert_equal Addressable::URI.parse('http://www.example.net/#foo'), JSON::Util::URI.normalize_ref("#foo", 'http://www.example.net')
   ```

2. Observe the failure and the incorrect URI being returned due to cache usage.

3. Clear the cache (`JSON::Util::URI.clear_cache`) between assertions to see the errors go away. 

You can find the problem reproduced in this PR https://github.com/voxpupuli/json-schema/pull/515

### Expected Behavior

The `JSON::Util::URI.normalize_ref` method should respect the `base` argument and resolve the reference correctly for each call.

### Actual Behavior

The method caches the parsed `ref` argument and incorrectly resolves it by ignoring the `base` argument on subsequent calls.

macfanatic:
Did #515 actually resolve this issue? I ask because the issue is still open, and that's unclear.

Came here from v5.0.0 release notes.

PR #515, opened 2024-07-05
Fix caching issues in JSON::Util::URI

bolshakov:
This PR fixes the issue #514

The implementation of the whole `JSON::Util::URI` was not thread-safe. For instance, if you look at the `.parse` method.

```ruby
def self.parse(uri)
  if uri.is_a?(Addressable::URI) 
    uri.dup
  else
    @parse_cache ||= {}
    parsed_uri = @parse_cache[uri]
    if parsed_uri
      parsed_uri.dup
    else
      @parse_cache[uri] = Addressable::URI.parse(uri)
    end
  end
rescue Addressable::URI::InvalidURIError => e
  raise JSON::Schema::UriError, e.message
end
```

In line `#8`, the URL is retrieved from the cache, and the caller receives a duplicate of the cached URL. However, in line `#10`, the caller does not receive a duplicate, and in a few instances, it changes the returned reference, effectively modifying it in the cache.  

```ruby
def self.normalize_ref(ref, base)
  ref_uri = parse(ref)
  base_uri = parse(base)

  ref_uri.defer_validation do
    if ref_uri.relative?
      ref_uri.merge!(base_uri)

      # Check for absolute path
      path, fragment = ref.to_s.split('#')
      if path.nil? || path == ''
        ref_uri.path = base_uri.path
      elsif path[0, 1] == '/'
        ref_uri.path = Pathname.new(path).cleanpath.to_s
      else
        ref_uri.join!(path)
      end

      ref_uri.fragment = fragment
    end

    ref_uri.fragment = '' if ref_uri.fragment.nil? || ref_uri.fragment.empty?
  end

  ref_uri
end
```

As you can see, if the referred URI is not in the cache, it is added to the cache and then modified inside the `.normalize_ref` method. Basically, we pass an object by reference between methods and modify it. This code is not thread-safe because it accesses and modifies a shared resource, `@parse_cache`, without any form of synchronization. 

Other methods that call `.parse` method to resolve references also implement a similar non-synchronized caching approach. In a multi-threaded environment, this could lead to a race condition and potential incorrect resolution of references in JSON Schemas.

Initially, I wanted to add Mutex synchronization or use thread-safe data structures (such as `Corcurrent::Map` for cache), but it appeared that the code is deeply intertwined, with many methods from the module calling each other, making synchronization prone to deadlocks.

I have removed caching from this module entirely because it is not thread-safe and causes more problems than it solves. The improvements it offers are minimal since we don't cache expensive network or disk calls but only in-memory calculations. 

I also fixed a couple of bugs in method implementations along the way:
*  Issue with the `URI.absolutize_uri` method. The `URI.absolutize_ref` implementation does not work as described in the tests. The tests pass only because of caching that persists state between subsequent calls in the test example.  See this PR showing how to reproduce the issue: https://github.com/voxpupuli/json-schema/pull/516. 
* Issue with the  `URI.normalized_ref` method which can persist state between calls leading to incorrect reference normalisation. This PR https://github.com/voxpupuli/json-schema/pull/517 reproduces the issue.

I tried to keep, changes in the tests suite at minimum, but ended up changing [this single test] which behaviour was not consistent with other test examples..

[this single test]: https://github.com/voxpupuli/json-schema/pull/515/files?diff=unified&w=0#diff-9337ccf2da90709cf54ecda66d8e961da93ed2410d4fe6ecf3f34d506a4b5143R130


bolshakov:
@ekohl, @TheMeier, @evgeni folks, may you also help with the review?

evgeni:
I have no idea about this gem, sorry.

bolshakov:
My apologies. I pinged you because you reviewed one of the recent PRs https://github.com/voxpupuli/json-schema/pull/509

lib/json-schema/util/uri.rb
```diff
@@ -1,109 +1,131 @@
+# frozen_string_literal: true
+
 require 'addressable/uri'
 
 module JSON
   module Util
-    module URI
+    class URI < Addressable::URI
```
bastelfreak, line 7: I'm wondering if this is a breaking change or not and I'm a bit unsure. What is your opinion @bolshakov ?

bolshakov: To be honest, I haven't thought of this helper as part of the public API. I would consider it rather private, to be used internally within the gem. The only change my PR introduces is that you cannot include URI as a module anymore since it's a class now. However, this feature is not documented in the README, so I wouldn't consider it part of the public API.

A good idea might be to add a comment explicitly stating this:

```suggestion
    # @api private 
    class URI < Addressable::URI
```

bastelfreak: I also don't think this is really a public API, but it might be used as one. I also want to do a major release to drop support for old rubies. So maybe we don't release this one as a Bugfix but with the next major release?

bolshakov: Your plan sounds good to me. Would you prefer adding `@api private` tag?

bastelfreak: I think that makes sense. Can you please add it?

bolshakov:
@bastelfreak ready

mjknight50:
Is there a timeline on when this will be merged?

bastelfreak:
Hi, I am sorry for the huge delay. Thanks for all the work and thanks for pinging me again!

PR #525, opened 2024-08-19
dev deps: Add strict version contraints & CI: build gem with strict mode & CI: Use Ruby 3.3 for releases

bastelfreak:
also contains #524

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,10 +11,10 @@ jobs:
     if: github.repository_owner == 'voxpupuli'
     steps:
       - uses: actions/checkout@v4
-      - name: Install Ruby 3.1
+      - name: Install Ruby 3.3
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: '3.1'
+          ruby-version: '3.3'
       - name: Build gem
         run: gem build --strict --verbose *.gemspec
       - name: Publish gem to rubygems.org
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,7 @@ jobs:
         with:
           ruby-version: '3.1'
       - name: Build gem
-        run: gem build *.gemspec
+        run: gem build --strict --verbose *.gemspec
       - name: Publish gem to rubygems.org
         run: gem push *.gem
         env:
--- a/json-schema.gemspec
+++ b/json-schema.gemspec
@@ -18,9 +18,9 @@ Gem::Specification.new do |s|
   s.license = 'MIT'
 
   s.add_development_dependency 'minitest', '~> 5.0'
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -11,12 +11,12 @@ jobs:
     if: github.repository_owner == 'voxpupuli'
     steps:
       - uses: actions/checkout@v4
-      - name: Install Ruby 3.1
+      - name: Install Ruby 3.3
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: '3.1'
+          ruby-version: '3.3'
```
ekohl, line 17: In another place I've been lazy and just used Ruby 3. Is that something we want here as well?
```suggestion
          ruby-version: '3'
```

json-schema.gemspec
```diff
@@ -16,13 +15,11 @@ Gem::Specification.new do |s|
   s.extra_rdoc_files = ['README.md', 'LICENSE.md']
   s.required_ruby_version = '>= 2.7'
   s.license = 'MIT'
-  s.required_rubygems_version = '>= 2.5'
 
-  s.add_development_dependency 'bundler'
   s.add_development_dependency 'minitest', '~> 5.0'
-  s.add_development_dependency 'rake'
+  s.add_development_dependency 'rake', '~> 13.2', '>= 13.2.1'
```
ekohl, line 20: Why so specific? I'd imagine this works just as well
```suggestion
  s.add_development_dependency 'rake', '~> 13.0'
```

json-schema.gemspec
```diff
@@ -16,13 +15,11 @@ Gem::Specification.new do |s|
...
   s.add_development_dependency 'voxpupuli-rubocop', '~> 2.8.0'
-  s.add_development_dependency 'webmock'
+  s.add_development_dependency 'webmock', '~> 3.23', '>= 3.23.1'
```
ekohl, line 22: Same here:
```suggestion
  s.add_development_dependency 'webmock', '~> 3.23'
```

json-schema.gemspec
```diff
@@ -16,13 +15,11 @@ Gem::Specification.new do |s|
...
 
-  s.add_runtime_dependency 'addressable', '>= 2.8'
+  s.add_runtime_dependency 'addressable', '~> 2.8', '>= 2.8.7'
```
ekohl, line 24:
```suggestion
  s.add_runtime_dependency 'addressable', '~> 2.8'
```

.github/workflows/release.yml
bastelfreak: I'm not sure. I don't know under which conditions they will bump that version. I only bump the version in the release action when we've the same version in a test and it passes.

TheMeier (approved):

PR #537, opened 2025-07-07
Make JSON::Validator freezable

hugopeixoto:
To avoid accidental memory leaks or other accidental modifications when reusing a validator object, a frozen validator object should still be able to validate schemas.

The only thing preventing us from freezing the validator object was the error tracking, which was temporarily stored in `@errors` and always cleaned up at the end of the #validate method.


bastelfreak:
thanks for the PR!

Issue #541, opened 2025-07-09
NameError: uninitialized constant JSON::Validator::SimpleDelegator (NameError)


ehelms:
After the 5.20 release, running `bundle exec rake` in a puppet module fails:

```
NameError: uninitialized constant JSON::Validator::SimpleDelegator (NameError)

    class ErrorRecorder < SimpleDelegator
                          ^^^^^^^^^^^^^^^
/home/ehelms/.rubygems/ruby/3.2.0/gems/json-schema-5.2.0/lib/json-schema/validator.rb:109:in `<class:Validator>'
/home/ehelms/.rubygems/ruby/3.2.0/gems/json-schema-5.2.0/lib/json-schema/validator.rb:18:in `<module:JSON>'
/home/ehelms/.rubygems/ruby/3.2.0/gems/json-schema-5.2.0/lib/json-schema/validator.rb:17:in `<top (required)>'
```

PR #542, opened 2025-07-09
Require delegate library in validator

beechtom:
This adds a missing require for the delegate library. Previously, a `NameError` was raised as the `SimpleDelegator` class could not be found.


bastelfreak:
fixes #541

PR #545, opened 2025-08-06
Improve :failed_attribute value for custom formats

bmdelaune:
When this failed_attribute param is parsed, it expects a class name, not an instance name. This will ensure it shows `CustomFormat` instead of `0x00009783abf73>`
image


bastelfreak:
@bmdelaune thanks for the fix! Is there any chance you can add a test for it?

jacobdaddario:
Very cool PR. Exactly what I needed for my project!

PR #568, opened 2026-02-11
Remove multi_json support

jordan-brough:
Adding multi_json to a project for an unrelated reason would silently cause json-schema to use it instead of the json gem, introducing subtle behavioral differences.

With modern Ruby and the json gem, multi_json is no longer needed.

See https://github.com/voxpupuli/json-schema/issues/565

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,10 +474,6 @@ validation:
 JSON::Validator.json_backend = :json
 ```
 
-Optionally, the JSON Schema library supports using the MultiJSON library for
-selecting JSON backends. If the MultiJSON library is installed, it will be
-autoloaded.
-
 Notes
 -----
 
--- a/lib/json-schema.rb
+++ b/lib/json-schema.rb
@@ -1,13 +1,5 @@
 require 'rubygems'
 
-if Gem::Specification.find_all_by_name('multi_json').any?
-  require 'multi_json'
-
-  # Force MultiJson to load an engine before we define the JSON constant here; otherwise,
-  # it looks for things that are under the JSON namespace that aren't there (since we have defined it here)
-  MultiJson.respond_to?(:adapter) ? MultiJson.adapter : MultiJson.engine
-end
-
 require 'json-schema/util/array_set'
 require 'json-schema/util/uri'
 require 'json-schema/schema'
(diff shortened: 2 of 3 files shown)


ekohl (approved):

ekohl:
@bastelfreak asked whether this really is backwards incompatible or not. If it is, we should probably do a release with the current unreleased changes.

thomaswitt:
Hi all — landing here from a real Rails 8 app where this PR is the only thing that silences a boot-time deprecation.

`multi_json` 1.21 turned the legacy `MultiJson` constant into a one-shot deprecation forwarder that fires when `lib/json-schema.rb:8` (`MultiJson.respond_to?(:adapter) ? MultiJson.adapter : MultiJson.engine`) runs at *require* time — before any consumer-set `JSON::Validator.use_multi_json = false` (from #566) can take effect. So the runtime opt-out is great, but doesn't silence the load-time noise.

On the backwards-compat concern that paused this thread: `multi_json` 1.21 keeps both `MultiJson` (deprecated forwarder) and `MultiJSON` (canonical) constants available, and similarly keeps `load`/`dump` as deprecated aliases for `parse`/`generate`. So a `defined?(::MultiJSON) ? ::MultiJSON : ::MultiJson` (plus `respond_to?(:generate) && respond_to?(:parse)` for the method names) switch silences every deprecation without bumping the gemspec floor or breaking any existing user contract on `multi_json ~> 1.15`. I shipped exactly that pattern in rgeo/rgeo-geojson#64 yesterday if a concrete reference helps — it's a 5-file diff with a regression test, fully backwards-compatible.

That said, the maintainer agreement upthread (@ekohl, @bastelfreak) is clearly to remove `multi_json` entirely, which is the cleaner long-term call. Happy to rebase this branch on top of current master and resolve the conflicts with credit to @jordan-brough if that unblocks things — just say the word.

(Mostly filing this to break the silence on the bastelfreak ping above; totally understand if waiting for a major version bump is the preferred path.)

spbever:
Just curious of the status of this update.  This is the last update we need to remove the deprecation warning from our app.  Anything I can do to assist?

PR #578, opened 2026-08-12
Stop passing `quirks_mode: true` to JSON.parse

byroot:
This option has been removed from the json gem in version 2.0.0 back in 2015.

Now with JSON 3.0.0.rc1, passing non-existing options raises `ArgumentError`.


smortex (approved):

yxhuvud:
And now that json 3.0.0 is actually released, this needs a release very soon or else all users will have to restrict the json gem version in their lock files.

